PR #1, opened 2020-08-11
Add a link to login and logout

sashapff:


src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,21 @@
 package com.google.alpollo;
 
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+
```
jeinygroove: nit: remove empty line between imports

src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,21 @@
...
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
-import com.google.appengine.api.users.UserService;
 
 @WebServlet("/auth")
 public class AuthenticationServlet extends HttpServlet {
     @Override
     protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         UserService userService = UserServiceFactory.getUserService();
+        String loginUrl = userService.createLoginURL("/");
```
jeinygroove: From this servlet you return login url, but on the frontend you use it as "Login | Logout" url, it'll be better to return json response with some information about user, if he's logged in and logout url, or logout url if not, so we'll know exactly is the user authenticated or not.

src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,21 @@
...
+        // Redirect to the login page and then to the main page
+        response.sendRedirect(loginUrl);
```
jeinygroove: We also won't force users to login, if they don't want to, so there's no need to redirect them to the login page.

src/main/webapp/index.html
```diff
@@ -11,5 +11,6 @@
           document.querySelector('p').innerText = text;
       });
   </script>
+  <a href="/auth">Login | Logout</a>
```
jeinygroove: It's better to add all tags in `<body>` before `<script>` tag, because if your js file will need some DOM element after `<script>` tag, it can not to find it.

src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,35 @@
...
-import com.google.appengine.api.users.UserService;
+import com.google.gson.Gson;
 
 @WebServlet("/auth")
 public class AuthenticationServlet extends HttpServlet {
+    private final Gson gson = new Gson();
+
+    private class LoggedInInfo {
+        final boolean loggedIn;
+        final String loginUrl;
```
jeinygroove: this field can also mean logoutUrl, maybe it's better to name it authUrl (I have no better idea).

src/main/webapp/index.html
```diff
@@ -6,10 +6,21 @@
...
 <body>
   <p></p>
+  <a></a>
   <script>
       fetch('/hello', {method: 'GET'}).then(response => response.text()).then((text) => {
           document.querySelector('p').innerText = text;
       });
+
+      fetch('/auth', {method: 'GET'}).then(response => response.json()).then((loggedInInfo) => {
+          let loginLogout = document.querySelector('a');
```
jeinygroove: it can be better to rename it to `authLinkElement`, because it says that this variable is and HTML Element (tag <a>) and this element provides authentication link (I found that authentication can be used as common word for login/logout)

jeinygroove (approved):

src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,37 @@
...
 import java.io.IOException;
-import com.google.appengine.api.users.UserService;
+import com.google.gson.Gson;
 
+/** Servlet that returns an auth info that includes status whether the user logged in and link to auth. */
 @WebServlet("/auth")
 public class AuthenticationServlet extends HttpServlet {
+    private final Gson gson = new Gson();
+
+    /** Class provides info about logged in status of the user and link to the authentication. */
```
amaurymedeiros: nit: "Class that provides [...]"

src/main/java/com/google/alpollo/AuthenticationServlet.java
```diff
@@ -1,16 +1,37 @@
...
+    private class LoggedInInfo {
+        private final boolean loggedIn;
```
amaurymedeiros: nit: it's a good practice to prefix boolean parameters with `has` or `is` whenever it makes sense. I think `isLoggedIn` makes sense here.

src/main/webapp/index.html
```diff
@@ -6,10 +6,21 @@
...
 <body>
   <p></p>
+  <a></a>
   <script>
       fetch('/hello', {method: 'GET'}).then(response => response.text()).then((text) => {
           document.querySelector('p').innerText = text;
       });
+
+      fetch('/auth', {method: 'GET'}).then(response => response.json()).then((loggedInInfo) => {
+          let authLinkElement = document.querySelector('a');
```
amaurymedeiros: Selecting by the anchor element is dangerous here. You'll probably have more `<a>` elements in your code later. It's better  to add an ID to the element and use it for the selector.

src/main/webapp/index.html
```diff
@@ -6,10 +6,21 @@
 </head>
 <body>
   <p></p>
+  <a></a>
   <script>
       fetch('/hello', {method: 'GET'}).then(response => response.text()).then((text) => {
           document.querySelector('p').innerText = text;
```
amaurymedeiros: Same comment as below. Please add an id to the `<p>` element and use it to select the element.

src/main/webapp/index.html
```diff
@@ -6,10 +6,21 @@
 </head>
```
amaurymedeiros, line 6: Can you publish the app so I can see the changes in go/alpollo-app?

amaurymedeiros (approved):

PR #5, opened 2020-08-17
Top songs page

jeinygroove:
This page was deployed, so you can see it [there](https://alpollo-step-2020.googleplex.com/)


src/main/webapp/src/components/RatingList.js
```diff
@@ -0,0 +1,105 @@
...
+        secondary={item.bandName + '. From album "' + item.albumName + '"'}
+      />
+    </ListItem>
+  ));
+  return <List>{reactListItems}</List>;
+}
+
+export default function RatingList() {
+  const classes = useStyles();
+
```
RazvanBudacaTechquarter: I think you should leave a comment telling people we will add songs automatically later.
Something like
// TODO Fetch top songs from database.

RazvanBudacaTechquarter (approved):
LGTM

sashapff (approved):

src/main/webapp/src/App.css
```diff
@@ -0,0 +1,11 @@
+@font-face {
```
amaurymedeiros: CSS and JS file names should be lowercase

src/main/webapp/public/index.html
```diff
@@ -25,19 +25,10 @@
       Learn how to configure a non-root public URL by running `npm run build`.
```
amaurymedeiros, line 25: Same comment as the other PR. Let's make sure the authentication PR doesn't get removed when this one is merged.

src/main/webapp/src/components/Navbar.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <div className={classes.root}>
+      <AppBar style={{ background: "#733F94" }} position="static">
+        <Toolbar>
+          <Typography variant="h6" className={classes.title}>
+            Alpollo
+          </Typography>
+          <Button color="inherit">Search</Button>
+          <Button color="inherit">Login</Button>
```
amaurymedeiros: Probably the authentication is going to be linked here... Maybe sync with @sashapff to implement it?

src/main/java/com/google/alpollo/Song.java
```diff
@@ -0,0 +1,29 @@
...
+  private final String album;
+
+  public Song(String artist, String name, String album) {
+    this.artist = artist;
+    this.name = name;
+    this.album = album;
+  }
+
+  /** Returns artists's name of the song. */
+  public String getArtists() {
```
jgaillard85: It is usually better to have field names matching their getter method. So I would rename this to getArtist, or rename the field artists

jgaillard85 (commented):
This PR is long and does several things. For the future, let's try and have each PR being the smallest block possible that makes sense logically

PR #6, opened 2020-08-17
Create SongInfo class

RazvanBudacaTechquarter:
This class will be used to store all the results of the analysis.

For now, it's just gonna store the name, score and entityList resulted from the analysis for easier and more elegant use.
In the future it could include albumName, genre, lyrics.


RazvanBudacaTechquarter:
@googlebot I signed it!

PR #8, opened 2020-08-18
Budi analyzes sentiments

RazvanBudacaTechquarter:
Sending a POST request to the servlet with the lyrics as parameter will return a JSON object with sentiment score and magnitude.

The AnalysisTest.java file was created so I can test if the analysis actually works, because currently there is no way to access the servlet from the site.

Sensitive data will be hidden later together with Olya.


jeinygroove:
Upload changes from master to this branch, please, because it has very old code.

pom.xml
```diff
@@ -33,6 +33,11 @@
             <artifactId>appengine-api-1.0-sdk</artifactId>
             <version>1.9.59</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-language</artifactId>
+            <version>1.55.0</version>
+        </dependency>
```
jeinygroove, line 40: I couldn't run this code, because compiled didn't see junit library, add dependency for it

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+  public Sentiment getSentiment(String lyrics) throws IOException{
+    Document doc =
+        Document.newBuilder().setContent(lyrics).setType(Document.Type.PLAIN_TEXT).build();
+
+    // Set the header manually so we can use the Natural Language API.
+    LanguageServiceSettings settings = LanguageServiceSettings.newBuilder().setHeaderProvider(
+        FixedHeaderProvider.create("X-Goog-User-Project","google.com:alpollo-step-2020")).build();
+    LanguageServiceClient languageService = LanguageServiceClient.create(settings);
+    Sentiment sentiment = languageService.analyzeSentiment(doc).getDocumentSentiment();
+    languageService.close();
```
jeinygroove: it's better to implement with try-resources statement, so you won't need to call close for this instance

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+public class AnalysisHelper {
+  
+  /**
+   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
+   * This sentiment has a score, showing the overall positivity of the text, ranging from -1 to 1
+   * and a magnitude, representing how strong the sentiment is, ranging from 0 to 1
+   */
+  public Sentiment getSentiment(String lyrics) throws IOException{
```
jeinygroove: space after `IOException`

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * Helper class for cleaner code and easier testing.
+ */
+
```
jeinygroove: nit: remove emply line

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
+package com.google.alpollo;
+
+import java.io.IOException;
+
```
jeinygroove: nit: remove empty line between imports

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -0,0 +1,29 @@
+package com.google.alpollo;
+
+import java.io.IOException;
+
```
jeinygroove: nit: remove empty line between imports

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -0,0 +1,55 @@
+package com.google.alpollo;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
```
jeinygroove: nit: remove empty line between imports

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+
```
jeinygroove: nit: remove empty line between imports

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * Helper class for cleaner code and easier testing.
```
jeinygroove: Please be more specific here, explain what this class, its methods do. I think there's no need to tell that it's "cleaner code and easier testing" :)

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+public class AnalysisHelper {
+  
+  /**
+   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
+   * This sentiment has a score, showing the overall positivity of the text, ranging from -1 to 1
+   * and a magnitude, representing how strong the sentiment is, ranging from 0 to 1
```
jeinygroove: Magnitude changes from `0.0` to `+inf`.

From the documentation: 
`magnitude indicates the overall strength of emotion (both positive and negative) within the given text, between 0.0 and +inf`

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Helper class for cleaner code and easier testing.
+ */
+
+public class AnalysisHelper {
+  
+  /**
+   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
+   * This sentiment has a score, showing the overall positivity of the text, ranging from -1 to 1
```
jeinygroove: nit: maybe I'm too strict here, but I think it's better to change `-1` and `1` to `-1.0` and `1.0`, because it'll correlate with documentation and it'll be obvious that we mean float number.

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -0,0 +1,15 @@
+package com.google.alpollo.model;
+
+/**
+ * Class used to store only sentiment score and magnitude.
+ * It will be easier to send necessary info to front-end as a JSON.
+ */
+public class SongSentiment {
```
jeinygroove, line 8: Why this class doesn't have getters/setters? You can write in documentation that this class is used only for converting its instances to json strings or smth like that.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,34 @@
...
+   * and a magnitude, representing how strong the sentiment is, ranging from 0 to 1
+   */
+  public Sentiment getSentiment(String lyrics) throws IOException{
+    Document doc =
+        Document.newBuilder().setContent(lyrics).setType(Document.Type.PLAIN_TEXT).build();
+
+    // Set the header manually so we can use the Natural Language API.
+    LanguageServiceSettings settings = LanguageServiceSettings.newBuilder().setHeaderProvider(
+        FixedHeaderProvider.create("X-Goog-User-Project","google.com:alpollo-step-2020")).build();
```
sashapff: nit: space between arguments in create method

sashapff, line 28: nit: space between arguments in create method

src/main/java/com/google/alpollo/model/SongSentiment.java
RazvanBudacaTechquarter: Well, isn't that exactly what it says? Should I rephrase it?

jeinygroove: Maybe you meant it in "store only sentiment score and magnitude", I didn't get it. Yes, I think it's better to rephrase it.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * Helper class with methods that use the Natural Language API or use the data that comes
+ * from the API
+ */
+public class AnalysisHelper {
+
```
sashapff, line 18: nit: remove empty line

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,29 @@
...
+   * and a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf
+   */
+  public Sentiment getSentiment(String lyrics) throws IOException {
+    LanguageServiceSettings settings = LanguageServiceSettings.newBuilder().setHeaderProvider(
+        FixedHeaderProvider.create("X-Goog-User-Project", "google.com:alpollo-step-2020")).build();
+    try (LanguageServiceClient language = LanguageServiceClient.create(settings)) {
+      Document doc = Document.newBuilder().setContent(lyrics).setType(Document.Type.PLAIN_TEXT).build();
+      Sentiment sentiment = language.analyzeSentiment(doc).getDocumentSentiment();
+      return sentiment;
+   }
```
jeinygroove: nit: fix indent here, move `}` to the right
And does it compile, when return is inside try block?

jeinygroove (approved):

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Yep, I'm amazed too that it compiles.
I mean, it worked when running mvn test.

jeinygroove: Fair enough, I see logic in it, but anyway looks unnatural :)

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class AnalysisHelper {
+  /**
+   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
+   * This sentiment has a score, showing the overall positivity of the text, ranging from -1.0 to 1.0
+   * and a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf
+   */
+  public Sentiment getSentiment(String lyrics) throws IOException {
+    LanguageServiceSettings settings = LanguageServiceSettings.newBuilder().setHeaderProvider(
+        FixedHeaderProvider.create("X-Goog-User-Project", "google.com:alpollo-step-2020")).build();
```
jgaillard85: Does this need to be re-created every time you use it, or could it be stored?

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,29 @@
...
+ * Helper class with methods that use the Natural Language API or use the data that comes
+ * from the API
+ */
+public class AnalysisHelper {
+  /**
+   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
+   * This sentiment has a score, showing the overall positivity of the text, ranging from -1.0 to 1.0
+   * and a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf
+   */
+  public Sentiment getSentiment(String lyrics) throws IOException {
```
jgaillard85: make the method static, that way you won't have to create an AnalysisHelper object to call it

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -0,0 +1,53 @@
...
+    helper = new AnalysisHelper();
+  }
+
+  @Test
+  public void checkSentimentScore() throws IOException {
+    Sentiment sentiment = helper.getSentiment(LYRICS);
+    float actual = sentiment.getScore();
+    float expected = -0.1f;
+    
+    Assert.assertEquals(expected, actual, 0.05f);
```
jgaillard85: You could extract the 0.05f as a constant if you are always going to use the same value for your asserts

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.IOException;
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * Helper class with methods that use the Natural Language API or use the data that comes
+ * from the API
```
amaurymedeiros: nit: finish the sentences of this and other javadocs with `.`

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.cloud.language.v1.Sentiment;
+import com.google.gson.Gson;
+
+@WebServlet("/sentiment")
+public class SentimentServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String lyrics = request.getParameter("lyrics");
```
amaurymedeiros: maybe extract "lyrics" to a constant?

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.alpollo.model.SongSentiment;
+import com.google.cloud.language.v1.Sentiment;
+import com.google.gson.Gson;
+
+@WebServlet("/sentiment")
+public class SentimentServlet extends HttpServlet {
```
amaurymedeiros, line 17: Please document the servlet

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -0,0 +1,16 @@
+package com.google.alpollo.model;
+
+/**
+ * Class created to store only sentiment score and magnitude, leaving the other useless properties
+ * of the Sentiment object.
+ * Class is mainly used for converting its instances to JSON strings.
+ */
+public class SongSentiment {
+  private final float score;
+  private final float magnitude;
```
amaurymedeiros: Please document the fields so we know what `score` and `magnitude` are

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: You're right, it would be better to store it, but when I tried to store it outside the method I got an error message like `Default constructor cannot handle exception type IOException`, so I left it like this until now.

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: So can I make all the methods static in this class and use them without creating an object?

src/main/java/com/google/alpollo/SentimentServlet.java
RazvanBudacaTechquarter: I'm not sure I can make the variable static, it gives me the following error:
`Illegal modifier for parameter LYRICS; only final is permitted`
From my understanding, variables can be made static only outside of methods, but in my case i have to use the `request` parameter.

amaurymedeiros: Yea, you need to define it at class-level (field), not method-level.

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Amazing, totally forgot about it.

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: I would have to handle the exception somehow at class-level?

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -0,0 +1,31 @@
...
+ * the overall positivity of the song and how strong the sentiment of the song is.
+ */
+@WebServlet("/sentiment")
+public class SentimentServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+  private static String LYRICS;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    LYRICS = request.getParameter("lyrics");
```
amaurymedeiros: Sorry, I was not clear... what I meant was having something like:

```
private static final String LYRICS_PARAM = "lyrics";

...
AnalysisHelper.getSentiment(request.getParameter(LYRICS_PARAM));
...
```

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -0,0 +1,31 @@
...
+ * Sending a POST request with the lyrics of the song as a parameter will give the Client
+ * the overall positivity of the song and how strong the sentiment of the song is.
+ */
+@WebServlet("/sentiment")
+public class SentimentServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+  private static String LYRICS;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
amaurymedeiros, line 22: The servlet shouldn't throw an `IOException`. Use a `try/catch` instead and in the `catch` block you send an error in the response

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -0,0 +1,18 @@
+package com.google.alpollo.model;
+
+/**
+ * Class created to store only sentiment score and magnitude, leaving the other useless properties
+ * of the Sentiment object.
+ * Class is mainly used for converting its instances to JSON strings.
+ */
+public class SongSentiment {
+  // Each sentiment has a score, showing the overall positivity of the text, ranging from -1.0 to 1.0.
```
amaurymedeiros: Use javadoc notation (i.e. `/**`) instead of inline comments (i.e. `//`) to document fields as well

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -0,0 +1,54 @@
...
+      + "I will never be the same\n" + "Mirror-mirror tell the truth of the old ones and the youth\n"
+      + "Of the things we hide deep from ourselves\n" + "Mirror-mirror show me now what will I become and how\n"
+      + "For now I'm just a mountain\n" + "I'm the mountain\n" + "Down the road that takes me to the Headley Grange\n"
+      + "I see a figure of a young man\n" + "He's torn with doubts, mistakes, his selfishness and rage\n"
+      + "But doing all the best he can\n" + "I'm not so blind to see\n" + "That this young man is me";
+
+  @Before
+  public void setUp() {
+    helper = new AnalysisHelper();
+  }
```
amaurymedeiros: You probably don't need that anymore since `getSentiment` is now static

src/main/java/com/google/alpollo/SentimentServlet.java
RazvanBudacaTechquarter: Olya did this already in her separate branch, so when we merge it should be ok. Should I do it too though?

amaurymedeiros: In that case I'm OK with leaving this change for the next PR.

src/main/java/com/google/alpollo/AnalysisHelper.java
jgaillard85: You could have a static field to store this. In the method, you check whether the field is null. If it is not null, use the value stored in it. Otherwise create a new LanguageServiceSettings and store the value in the filed for the future.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.language.v1.Sentiment;
+
+/**
+ * Helper class with methods that use the Natural Language API or use the data that comes
+ * from the API.
+ */
+public class AnalysisHelper {
```
jgaillard85, line 14: If this class is just for static helper methods, create a private constructor for it so that it can never be instantiated by mistake

RazvanBudacaTechquarter: Already ahead of you ;)

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -0,0 +1,35 @@
...
+  private static LanguageServiceSettings SETTINGS;
```
jgaillard85: This is not a constant as it is not final, use lower case for the name

RazvanBudacaTechquarter: You're right, I confused static with constant.

amaurymedeiros (approved):

PR #9, opened 2020-08-19
Move sensitive data to config.json

jeinygroove:
Maybe it's better to retrieve Json file in the static block of the class, but in this case we won't be able to handle exceptions from it, on the other hand we won't need to generate this json object in each method. @sashapff, @BudiGabe what do you think?


src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -1,31 +1,33 @@
...
-  
+  private final Gson gson = new Gson();
+
   /**
    * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
    * This sentiment has a score, showing the overall positivity of the text, ranging from -1 to 1
    * and a magnitude, representing how strong the sentiment is, ranging from 0 to 1
    */
-  public Sentiment getSentiment(String lyrics) throws IOException{
+  public Sentiment getSentiment(String lyrics) throws Exception {
```
sashapff: I'm not sure that using all kinds of exceptions here is good idea. Specify which ones you want to throw.

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Objects;
+
+/**
+ * Helps to load configuration file and retrieve data from it.
+ */
+final public class ConfigHelper {
+    private static final String configFileName = "config.json";
```
sashapff: nit: change the indent to 2 spaces in this class

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -1,31 +1,33 @@
...
     Document doc =
-        Document.newBuilder().setContent(lyrics).setType(Document.Type.PLAIN_TEXT).build();
+            Document.newBuilder().setContent(lyrics).setType(Document.Type.PLAIN_TEXT).build();
```
sashapff: nit: change the indent to 4 spaces

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -1,31 +1,33 @@
...
+
+    String projectID = ConfigHelper.getProjectID();
 
     // Set the header manually so we can use the Natural Language API.
     LanguageServiceSettings settings = LanguageServiceSettings.newBuilder().setHeaderProvider(
-        FixedHeaderProvider.create("X-Goog-User-Project","google.com:alpollo-step-2020")).build();
+            FixedHeaderProvider.create("X-Goog-User-Project", projectID)).build();
```
sashapff: nit: change the indent to 4 spaces

src/test/resources/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "projectID": "your project ID"
```
sashapff, line 2: Is it okay to use "your project ID" here?

src/test/java/com/google/alpollo/SentimentTest.java
```diff
@@ -36,7 +34,7 @@ public void setUp() {
   }
 
   @Test
-  public void checkSentimentScore() throws IOException {
+  public void checkSentimentScore() throws Exception {
```
sashapff: The same with exceptions

src/test/java/com/google/alpollo/SentimentTest.java
```diff
@@ -45,7 +43,7 @@ public void checkSentimentScore() throws IOException {
   }
 
   @Test
-  public void checkSentimentMagnitude() throws IOException {
+  public void checkSentimentMagnitude() throws Exception {
```
sashapff: The same with exceptions

src/main/resources/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "projectID": "your project ID"
```
sashapff, line 2: Is it okay to use "your project ID" here?

src/test/resources/config.json
jeinygroove: yes, it's just default file, so you won't need to create file and folder for it manually, you need only change "your project ID" with project id

src/main/resources/config.json
jeinygroove: yes, it's just default file, so you won't need to create file and folder for it manually, you need only change "your project ID" with project id

jeinygroove: I can change it to `""` and add a comment btw, but I don't think that it's equivalent

sashapff: Ok, I just wanted to make sure

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -18,12 +20,19 @@
...
     AnalysisHelper helper = new AnalysisHelper();
-    Sentiment sentiment = helper.getSentiment(lyrics);
-    SongSentiment songSentiment = new SongSentiment(sentiment.getScore(), sentiment.getMagnitude());
-    
-    String json = gson.toJson(songSentiment);
-    response.setContentType("application/json;");
-    response.getWriter().println(json);
-  }
+    try {
+      Sentiment sentiment = helper.getSentiment(lyrics);
```
BudiGabe, line 24: I think it's better to move the try {...} inside the getSentiment() method, like I did on my PR. I don't consider it necessary for all the lines of code inside this doPost method.

sashapff (approved):

src/main/java/com/google/alpollo/SentimentServlet.java
jeinygroove: It's not possible because in catch block I send the error in response.

RazvanBudacaTechquarter (approved):

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -6,24 +6,33 @@
...
+import com.google.gson.Gson;
+import com.google.gson.JsonIOException;
+import com.google.gson.JsonSyntaxException;
 
 /**
  * Helper class with methods that use the Natural Language API or use the data that comes
  * from the API
  */
 public class AnalysisHelper {
+  private final Gson gson = new Gson();
```
amaurymedeiros: You don't seem to be using it, so there is no need to add it

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,41 @@
...
+    public String getProjectID() {
+      return projectID;
+    }
+  }
+
+  private ConfigHelper() {
+    throw new RuntimeException("Instantiation of ConfigHelper is not allowed!");
+  }
+
+  public static String getProjectID() throws NullPointerException,
```
amaurymedeiros: You don't need to declare `RuntimeException`s, e.g. `NullPointerException`. They fail at run time, so there is no need to tell the compiler about them.

Also, you should not simply propagate these JSON exceptions, otherwise all the callers will need to declare/import them as well. Instead, you can have a `try/catch` here and do one of the following when catching them:

- Throw another exception with a relevant message, e.g. "Failed to obtain Project ID.". The exception thrown can be an `IllegalStateException` or another one from the Java standard library that you think is more appropriate. The caller can then use a try/catch or rethrow the Exception without needing to import custom exceptions, e.g. `JsonIOException`.
- Return null. In this case the caller will check if the `projectId` is null and respond accordingly.

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -17,11 +19,19 @@
...
+      String json = gson.toJson(songSentiment);
+      response.setContentType("application/json;");
+      response.getWriter().println(json);
+    } catch (JsonIOException | JsonSyntaxException jsonException){
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+              "Json file with configuration info is invalid");
+    } catch (Exception e) {
+      response.sendError(HttpServletResponse.SC_NOT_FOUND,
+              "Internal error occurred or config.json file not found");
+    }
```
amaurymedeiros: This `SentimentServlet` doesn't need to know details about the errors that can happen in the `ConfigHelper`.  `getSentiment` should only throw one type of exception (maybe `IllegalStateException`), and then you can catch it here and return a single `response.sendError`, which can reuse the exception message.

src/test/resources/config.json
amaurymedeiros: That's OK. An alternative would be giving an example as well. For instance: `"your project ID, e.g. my-gcp-project-123"`

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -17,11 +17,16 @@
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
amaurymedeiros, line 22: ideally the `doPost` method shouldn't throw an exception. It should handle all the Exceptions in a try/catch block and return a proper error like you're doing in lines 27-30. You can probably remove the `throws` declaration here. If there is another source that can throw this exception, it should be handled as well

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,48 @@
...
+   * Retrieves projectID from configuration file.
+   * @return projectID; {@code null} if configuration file wasn't found
+   * or it was incorrect/didn't have projectID field
+   */
+  public static String getProjectID() {
+    try {
+      InputStream inputStream = ClassLoader.getSystemResourceAsStream(configFileName);
+      final Reader reader = new InputStreamReader(Objects.requireNonNull(inputStream));
+      return gson.fromJson(reader, ConfigInfo.class).getProjectID();
+    } catch (NullPointerException | JsonIOException | JsonSyntaxException parseException) {
```
amaurymedeiros: You can be generic here and just use `Exception parseException`. You need/want to return null if anything goes wrong anyway.

src/main/java/com/google/alpollo/SentimentServlet.java
jeinygroove: I can't simply remove it, because sendError method can throw IOException and I call this method in the catch block, looks like recursion 😆 

sendError method has `throws IOException`, so I think that's the reason why we can get rid of `ServletException`, but not of `IOException` while overriding `doPost`

Or there's still a way to do it?

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.Objects;
+
+/**
+ * Helps to load configuration file and retrieve data from it.
+ */
+final public class ConfigHelper {
+  private static final String configFileName = "config.json";
```
jgaillard85: It is customary to write constant names in all caps, so this one should be something like CONFIG_FILE_NAME

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * Retrieves projectID from configuration file.
+   * @return projectID; {@code null} if configuration file wasn't found
+   * or it was incorrect/didn't have projectID field
+   */
+  public static String getProjectID() {
+    try {
+      InputStream inputStream = ClassLoader.getSystemResourceAsStream(configFileName);
```
jgaillard85: Streams need to be closed, use the try with resources construct here

src/main/java/com/google/alpollo/SentimentServlet.java
```diff
@@ -21,11 +21,16 @@
...
-    response.setContentType("application/json;");
-    response.getWriter().println(json);
+    try {
+      Sentiment sentiment = AnalysisHelper.getSentiment(LYRICS);
+      SongSentiment songSentiment = new SongSentiment(sentiment.getScore(), sentiment.getMagnitude());
+
+      String json = gson.toJson(songSentiment);
+      response.setContentType("application/json;");
+      response.getWriter().println(json);
+    } catch (IllegalStateException | IOException sentimentException){
```
jgaillard85: add a space before the last {

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Retrieves projectID from configuration file.
+   * @return projectID; {@code null} if configuration file wasn't found
+   * or it was incorrect/didn't have projectID field
+   */
+  public static String getProjectID() {
+    try (InputStream inputStream = ClassLoader.getSystemResourceAsStream(CONFIG_FILE_NAME)) {
+      final Reader reader = new InputStreamReader(Objects.requireNonNull(inputStream));
```
jgaillard85: I should have been more specific, the reader also needs to be declared inside the try (), as it is also a Closeable

jeinygroove: sure, I forgot(

jgaillard85 (approved):

PR #11, opened 2020-08-19
Budi analyzes entities

RazvanBudacaTechquarter:
Sending a POST request to the servlet with the lyrics as parameter will return a list of the most salient entities, each having a name and a salience score.

Sensitive data will be hidden later with Olya, after the branches merge. 
Tests make the PRs look huge. Maybe I should separate them? If you have any test suggestions, feel free to message me.


src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -1,8 +1,16 @@
 package com.google.alpollo;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
```
jeinygroove: remove empty line between imports

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -17,13 +25,65 @@
...
+  /**
+   * From every entity generated by the API we get only part of the information
+   * and store the new SongEntities into a new simplified list. 
+   */
+  public static List<SongEntity> getSimplifiedEntityList(List<Entity> entityList) {
+    List<SongEntity> simplifiedEntityList = new ArrayList<>();
+
+    for (Entity entity : entityList) {
+      SongEntity simplifiedEntity = new SongEntity(entity.getName(), 
+          Math.round(entity.getSalience()*100.0)/100.0);
```
jeinygroove: add spaces in math operations

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -17,13 +25,65 @@
...
+      AnalyzeEntitiesRequest request = AnalyzeEntitiesRequest.newBuilder().setDocument(doc)
+          .setEncodingType(EncodingType.UTF16).build();
+      AnalyzeEntitiesResponse response = language.analyzeEntities(request);
+
+      return response.getEntitiesList();
+    }
+  }
+
+  /**
+   * From every entity generated by the API we get only part of the information
```
jeinygroove: it's better to specify what part of information do we get

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -5,13 +5,23 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+
```
jeinygroove: remove empty line between imports

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -5,13 +5,23 @@
...
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
```
jeinygroove: remove empty line between imports

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -17,13 +25,65 @@
...
+      simplifiedEntityList.add(simplifiedEntity);
+    }
+
+    return simplifiedEntityList;
+  }
+
+  public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
+    List<SongEntity> topSalientEntities = new ArrayList<>();
+    
+    if (simplifiedEntityList.size() <= 10) {
```
jeinygroove: I am sure that there must be a simple function for this action. I know that Scala has `.slice(10)`, for example, which takes first 10 elements of a list or all of them, it the length of the list is `<10`.

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
+package com.google.alpollo;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
```
jeinygroove: remove empty line between imports

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
```
jeinygroove: remove empty line between imports

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+  private static String LYRICS;
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    LYRICS = request.getParameter("lyrics");
+
+    // Get the Entity list from the API
+    List<Entity> entityList = new ArrayList<>(AnalysisHelper.getEntityList(LYRICS));
+
+    // Make a new list with simplified entities, they work better in the frontend
```
jeinygroove: I think that this comment can be deleted, we simplify entities not because it will be more comfortable to work with them on the client side, but because we don't need all the information which entity contains, that's why there's no need to send from servlet a lot of useless and more heavy data.

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public SongEntity(String name, double salience) {
+    this.name = name;
+    this.salience = salience;
+  }
+
+  /**
+   * A comparator for sorting song entities by their salience in ascending order.
+   */
+  public static final Comparator<SongEntity> ORDER_BY_SALIENCE = new Comparator<SongEntity>() {
```
jeinygroove: It would be great to specify the order in the name of the variable

jeinygroove: Also maybe it's better to make it descending, because you only use it with `reverseOrder`

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -44,10 +57,42 @@ public void checkSentimentScore() throws IOException {
...
-    Sentiment sentiment = helper.getSentiment(LYRICS);
+    Sentiment sentiment = helper.getSentiment(LYRICS_LONG);
     float actual = sentiment.getMagnitude();
     float expected = 0.7f;
     
     Assert.assertEquals(expected, actual, 0.05f);
   }
+
+  /**
+   * If the lyrics are too short, the API might find less than 10 entities.
```
jeinygroove, line 59: I'm not sure that we need to write tests which depends on the other API, because what if API will change and it'll send more than 10 entities in this text or it'll change an algorithm for defining salience of the word.

I suggest to emulate the result of the API, for example you can manually define `entityList`, so you'll check only your own methods.

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -44,10 +57,42 @@ public void checkSentimentScore() throws IOException {
...
+    String expected = gson.toJson(Arrays.asList(new SongEntity("mountain", 0.84), 
+        new SongEntity("mountain", 0.06), new SongEntity("story", 0.05), new SongEntity("sale", 0.03), 
+        new SongEntity("tale", 0.01)));
+        
+    Assert.assertEquals(expected, actual);
+  }
+
+  @Test
+  public void top10SalientEntitiesWithMoreThan10Entities() throws IOException {
+    List<Entity> entityList = new ArrayList<>(helper.getEntityList(LYRICS_LONG));
```
jeinygroove: I'm not sure that we need to write tests which depends on the other API, because what if API will change and it'll send more than 10 entities in this text or it'll change an algorithm for defining salience of the word.

I suggest to emulate the result of the API, for example you can manually define entityList, so you'll check only your own methods.

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -5,13 +5,23 @@
...
+
+import com.google.alpollo.model.SongEntity;
+import com.google.cloud.language.v1.Entity;
 import com.google.cloud.language.v1.Sentiment;
+import com.google.gson.Gson;
 
 @RunWith(JUnit4.class)
 public final class AnalysisTest {
+  private final Gson gson = new Gson();
   private AnalysisHelper helper;
```
jeinygroove: you call only static methods of the `AnalysisHelper`, so you don't need this variable, just call them with the class name

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -1,8 +1,16 @@
 package com.google.alpollo;
```
jeinygroove, line 2: Now I see that all methods in this class are `static`, so it's good to make it stateless

This's called Utility class and you can find a lot of best practises for it, for example, [here](https://stackoverflow.com/questions/31409982/java-best-practice-class-with-only-static-methods)

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -44,10 +57,42 @@ public void checkSentimentScore() throws IOException {
...
+    simplifiedEntityList.sort(Collections.reverseOrder(SongEntity.ORDER_BY_SALIENCE));
+
+    String actual = gson.toJson(helper.getTopSalientEntities(simplifiedEntityList));
+    String expected = gson.toJson(Arrays.asList(new SongEntity("mountain", 0.84), 
+        new SongEntity("mountain", 0.06), new SongEntity("story", 0.05), new SongEntity("sale", 0.03), 
+        new SongEntity("tale", 0.01)));
+        
+    Assert.assertEquals(expected, actual);
+  }
+
```
jeinygroove, line 73: Please, add comment to this test

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: I don't think there's any method like that in Java. I'll give it a search.

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: Initially I added everything because I wanted to check if everything works correctly. I could make a separate test to see if `getEntityList()` works and use a manually defined list here.
Let's see what the guys have to say about it too.

src/main/java/com/google/alpollo/AnalysisHelper.java
jeinygroove: [Look what I've found](https://stackoverflow.com/questions/11001720/get-only-part-of-an-array-in-java)

RazvanBudacaTechquarter: Apparently if the old array has less elements than you request, it will fill the new one with nulls.

jeinygroove: yes, but for `to` parameter you can make `int to = min(10, arr.size())`, problem solved)

jeinygroove: Also avoid "magic constants", move `10` to the constant variable

jeinygroove (commented):
I merged your branch with config changes.

Please, don't forget regularly upload new changes to your branch and fix conflicts.

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Tried using the array methods like this:
`List<SongEntity> topSalientEntities = Arrays.asList(Arrays.copyOfRange(simplifiedEntityList.toArray(), 0, 9));`

However, when converting back to list, it contains only simple objects, not SongEntities.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
...
+      simplifiedEntityList.add(simplifiedEntity);
+    }
+
+    return simplifiedEntityList;
+  }
+
+  public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
+    List<SongEntity> topSalientEntities = new ArrayList<>();
+    
+    if (simplifiedEntityList.size() <= 10) {
```
jeinygroove: `MAX_ENTITIES` instead of `10`

jeinygroove (approved):

jeinygroove:
@BudiGabe please, name commits according to the code, so we'll be able to understand what was done without opening it

RazvanBudacaTechquarter:
Okok no more jokes in commits :))

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
  * Helper class with methods that use the Natural Language API or use the data that comes
  * from the API.
  */
-public class AnalysisHelper {
+public final class AnalysisHelper {
   private static LanguageServiceSettings settings;
+  private static int MAX_ENTITIES = 10;
```
jgaillard85: Constants are static and final

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
...
-  
-  /**
-   * Based on the lyrics sent, the AI can extract the main "sentiment" of the text.
-   * This sentiment has a score, showing the overall positivity of the text, ranging from -1.0 to 1.0
-   * and a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf.
-   */
-  public static Sentiment getSentiment(String lyrics) throws IllegalStateException, IOException {
-    String projectID = ConfigHelper.getProjectID();
+
+  private static void getLanguageServiceSettings() throws IOException {
```
jgaillard85: The name of this method is misleading because it looks like a getter method but does not return anything.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
...
+  /**
+   * From every entity generated by the API we get only part of the information (name and salience)
+   * and store the new SongEntities into a new simplified list. 
+   */
+  public static List<SongEntity> getSimplifiedEntityList(List<Entity> entityList) {
+    List<SongEntity> simplifiedEntityList = new ArrayList<>();
+
+    for (Entity entity : entityList) {
+      SongEntity simplifiedEntity = new SongEntity(entity.getName(), 
+          Math.round(entity.getSalience() * 100.0) / 100.0);
```
jgaillard85, line 90: What is the point of this operation?

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
...
+      simplifiedEntityList.add(simplifiedEntity);
+    }
+
+    return simplifiedEntityList;
+  }
+
+  public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
+    List<SongEntity> topSalientEntities = new ArrayList<>();
```
jgaillard85: Only create this object if needed, so put it in the else branch

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,89 @@
...
+    
+    if (simplifiedEntityList.size() <= MAX_ENTITIES) {
+      return simplifiedEntityList;
+    } else {
+      for (int i = 0; i < MAX_ENTITIES; i++) {
+        topSalientEntities.add(simplifiedEntityList.get(i));
+      }
```
jgaillard85: You could use the sublist method to avoid creating a new list object for this, if you don't need a copy

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+import com.google.gson.Gson;
+
+/**
+ * Sending a POST request with the lyrics of the song as a parameter will return to the client
+ * a list of the most important entities (people, places, things) found in that song.
+ */
+@WebServlet("/entity")
+public class EntityServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+  private static String LYRICS;
```
jgaillard85: This is not a constant, so the field name should be lower case

jgaillard85: This doesn't even need to be a field, it could be a local variable

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      LYRICS = request.getParameter("lyrics");
+
+      // Get the Entity list from the API
+      List<Entity> entityList = new ArrayList<>(AnalysisHelper.getEntityList(LYRICS));
```
jgaillard85: Why do you create a copy of the list returned by getEntityList here?

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+      List<SongEntity> simplifiedEntityList = AnalysisHelper.getSimplifiedEntityList(entityList);
+      simplifiedEntityList.sort(SongEntity.ORDER_BY_SALIENCE_DESCENDING);
```
jgaillard85: Maybe this sorting should happen in the getTopSalientEntities, as the method name suggests it returns the top ones, while really at the moment it returns the first ones

src/test/java/com/google/alpollo/AnalysisTest.java
jgaillard85: I think you should have tests that test that your own code is doing what it is supposed to do, and also tests that make sure that the API you are relying on is still doing what you think it is doing. It is generally better to have those separate, so that you immediately know if something breaks whether it is something you did or it is related to a change in the API

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: I used this to format the double to always show 2 decimal places. It was pretty ugly with all those decimals.

src/main/java/com/google/alpollo/EntityServlet.java
RazvanBudacaTechquarter: Ohh that's my bad, apparently I forgot some stuff after merging. Thanks for the heads up,

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: Should I leave that to a separate PR, as it is already pretty huge?

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: You can store the value as it is. When displaying that to the user, i.e. printing to the UI, you can format the string to display only 2 decimal places

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -91,16 +93,13 @@ public static Sentiment getSentiment(String lyrics) throws IllegalStateException
   }
 
   public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
-    List<SongEntity> topSalientEntities = new ArrayList<>();
-    
+    simplifiedEntityList.sort(SongEntity.ORDER_BY_SALIENCE_DESCENDING);
```
amaurymedeiros: This modifies the list passed as an argument. We shouldn't do that as it's not expected by looking at the method signature. You should make a copy instead.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -91,16 +93,13 @@ public static Sentiment getSentiment(String lyrics) throws IllegalStateException
...
+
     if (simplifiedEntityList.size() <= MAX_ENTITIES) {
       return simplifiedEntityList;
     } else {
-      for (int i = 0; i < MAX_ENTITIES; i++) {
-        topSalientEntities.add(simplifiedEntityList.get(i));
-      }
+      List<SongEntity> topSalientEntities = simplifiedEntityList.subList(FIRST_ENTITY, MAX_ENTITIES);
+      return topSalientEntities;
```
amaurymedeiros: nit: you can merge these lines, e.g. `return simplifiedEntityList.subList(FIRST_ENTITY, MAX_ENTITIES);`

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -18,17 +18,16 @@
 @WebServlet("/entity")
 public class EntityServlet extends HttpServlet {
   private final Gson gson = new Gson();
-  private static String LYRICS;
+  private static String LYRICS = "lyrics";
```
amaurymedeiros: Document this constant saying it's a parameter from requests to the servlet, etc.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -20,12 +20,13 @@
  */
 public final class AnalysisHelper {
   private static LanguageServiceSettings settings;
-  private static int MAX_ENTITIES = 10;
+  private static final int MAX_ENTITIES = 10;
+  private static final int FIRST_ENTITY = 0;
```
amaurymedeiros: Please document these constants

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Yes, it can be fixed while printing the UI, but without rounding I can't really write tests for it. The online API only shows me the first 2 decimals, so the only way to compare them was to round the numbers.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -11,31 +18,90 @@
...
+  }
+
+  public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
+    List<SongEntity> tempList = new ArrayList<>(simplifiedEntityList);
+    tempList.sort(SongEntity.ORDER_BY_SALIENCE_DESCENDING);
+
+    if (tempList.size() <= MAX_ENTITIES) {
+      return tempList;
+    } else {
+        return simplifiedEntityList.subList(FIRST_ENTITY, MAX_ENTITIES);
```
jgaillard85: Here I think you want a sublist of tempList.

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Sending a POST request with the lyrics of the song as a parameter will return to the client
+ * a list of the most important entities (people, places, things) found in that song.
+ */
+@WebServlet("/entity")
+public class EntityServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+  /** Parameter sent through the request which holds the song's lyrics. */
+  private static String LYRICS = "lyrics";
```
jgaillard85: For it to be a constant, you have to make it final

src/test/java/com/google/alpollo/AnalysisTest.java
jgaillard85: You can leave it as is

jgaillard85 (approved):

PR #12, opened 2020-08-20
Create a class to complement Song.java

RazvanBudacaTechquarter:
This class will be used to store all the information taken from the server, then we are going to use it to store info about the most searched songs.

### DIFF
--- a/src/main/java/com/google/alpollo/SongInfo.java
+++ b/src/main/java/com/google/alpollo/SongInfo.java
@@ -7,7 +7,7 @@ import com.google.alpollo.model.SongEntity;
  * Class used for storing data related to the Natural Language API analysis.
  * Extending the Song class, it provides all the info related to a specific song.
  */
-public class SongInfo extends Song{
+public class SongInfo extends Song {
   /** Score tells us the overall positivty of the song, ranging from -1.0 to 1.0.*/
   private final float score;
   /** Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf. */
--- a/src/main/java/com/google/alpollo/SongInfo.java
+++ b/src/main/java/com/google/alpollo/SongInfo.java
@@ -1,0 +1,38 @@
+package com.google.alpollo;
+
+import java.util.List;
+import com.google.alpollo.model.SongEntity;
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ * 
+ * Score tells us the overall positivty of the song, ranging from -1.0 to 1.0.
+ * Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf.
+ * TopSalientEntities is a list containing the 10 most important words, given the song context.
+ */
+public class SongInfo extends Song{
+  private final float score;
+  private final float magnitude;
+  private final List<SongEntity> topSalientEntities;
+
+  public SongInfo(String artist, String name, String album, float score, float magnitude,
+      List<SongEntity> topSalientEntities) {
+    super(artist, name, album);
+    this.score = score;
+    this.magnitude = magnitude;
+    this.topSalientEntities = topSalientEntities;
+  }
+
+  public float getScore() {
+    return score;
+  }
+
+  public float getMagnitude() {
+    return magnitude;
+  }
+
+  public List<SongEntity> getTopSalientEntities() {
+    return topSalientEntities;
+  }
+}
--- a/src/main/java/com/google/alpollo/SongInfo.java
+++ b/src/main/java/com/google/alpollo/SongInfo.java
@@ -6,14 +6,13 @@ import com.google.alpollo.model.SongEntity;
 /**
  * Class used for storing data related to the Natural Language API analysis.
  * Extending the Song class, it provides all the info related to a specific song.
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ * 
+ * Score tells us the overall positivty of the song, ranging from -1.0 to 1.0.
+ * Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf.
+ * TopSalientEntities is a list containing the 10 most important words, given the song context.
+ */
+public class SongInfo extends Song{
```
jeinygroove: space before `{`

src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,38 @@
+package com.google.alpollo;
+
+import java.util.List;
+import com.google.alpollo.model.SongEntity;
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ * 
+ * Score tells us the overall positivty of the song, ranging from -1.0 to 1.0.
```
jeinygroove: move these comments to fields

src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,38 @@
+package com.google.alpollo;
+
+import java.util.List;
+import com.google.alpollo.model.SongEntity;
```
jeinygroove, line 4: SongEntity doesn't exist in this branch, code doesn't compile

jeinygroove (commented):
Seems like you wanted to create this branch from `BudiAnalysis`, but it's not

src/main/java/com/google/alpollo/SongInfo.java
RazvanBudacaTechquarter: Totally true, I should have branched from BudiEntityAnalisis, but everything should be fine if we merge that first.

src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ */
+public class SongInfo extends Song {
+  /** Score tells us the overall positivty of the song, ranging from -1.0 to 1.0.*/
```
jeinygroove, line 11: it can be just `//`

src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,37 @@
...
+  private final float score;
+  /** Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf. */
```
jeinygroove, line 13: it can be just `//`

src/main/java/com/google/alpollo/SongInfo.java
```diff
@@ -0,0 +1,37 @@
...
+  private final float magnitude;
+  /** TopSalientEntities is a list containing the 10 most important words, given the song context.*/
```
jeinygroove, line 15: it can be just `//`

src/main/java/com/google/alpollo/SongInfo.java
RazvanBudacaTechquarter: Hmm, the guys told me before I should use javadoc

jeinygroove:
And I still don't see `SongEntity`, please upload changes from `master` then. It's not good when code in your branch doesn't compile, for example, if you will merge this branch to master, it can stop compiling as well.

It's good that you already have `BudiAnalysis` merged to `master`, but if it wasn't, we could have the broken `master` after this merge :(

jeinygroove (approved):

src/main/java/com/google/alpollo/SongInfo.java
jeinygroove: OK

jgaillard85 (commented):
I am usually all for short PRs, but here you create a new class without using it. That makes it hard to see whether this is good or not. For example, I don't know at the moment why you are adding this class, whether having this as a subclass of Song is the best approach...

RazvanBudacaTechquarter:
> I am usually all for short PRs, but here you create a new class without using it. That makes it hard to see whether this is good or not. For example, I don't know at the moment why you are adding this class, whether having this as a subclass of Song is the best approach...

I agree with you, this class was created while keeping in mind future plans. Maybe it's better to leave it for later.

RazvanBudacaTechquarter:
> And I still don't see `SongEntity`, please upload changes from `master` then. It's not good when code in your branch doesn't compile, for example, if you will merge this branch to master, it can stop compiling as well.
> 
> It's good that you already have `BudiAnalysis` merged to `master`, but if it wasn't, we could have the broken `master` after this merge :(

SongEntity is part of the BudiEntityAnalysis branch, which is not yet merged. This feature should have been branched from BudiEntityAnalysis.

PR #13, opened 2020-08-20
Add routing to song page with song info

jeinygroove:
You can see this page by adding `/song` to the URL. 
Or use the [link](https://alpollo-step-2020.googleplex.com/#/song)


src/main/webapp/src/app.js
```diff
@@ -1,7 +1,13 @@
...
+import SongInfo from "./components/songInfo";
 import axios from "axios";
+import {
+  HashRouter as Router,
+  Switch,
+  Route
+} from "react-router-dom";
 import "./app.css";
 
 class App extends Component {
```
RazvanBudacaTechquarter, line 17: Please document this class.

src/main/webapp/src/components/songInfo.js
```diff
@@ -0,0 +1,205 @@
...
+    width: "640px",
+    margin: "10px",
+    listStyleType: "none",
+  },
+  languageAnalysisSection: {
+    display: "flex",
+  },
+});
+
+class SongInfo extends React.Component {
```
RazvanBudacaTechquarter, line 53: Please document this class.

RazvanBudacaTechquarter (approved):
Ok so my main concern is that all the code (HTML, CSS and JS) is dumped in a single file, but I suppose that's how React is done? Maybe separate everything neatly in different files if possible.

jeinygroove:
> Ok so my main concern is that all the code (HTML, CSS and JS) is dumped in a single file, but I suppose that's how React is done? Maybe separate everything neatly in different files if possible.

Yes, as I told you before I'm planning to do smth like that, at this moment I'm looking for the best way to do it

src/main/webapp/src/components/songInfo.js
```diff
@@ -0,0 +1,208 @@
...
+        },
+        {
+          word: "selling show",
+          salience: "0.04",
+          category: "work of art",
+        },
+        {
+          word: "God",
+          salience: "0.01",
+          category: "person",
```
amaurymedeiros, line 130: That's interesting, lol

src/main/webapp/src/components/songInfo.js
```diff
@@ -0,0 +1,208 @@
...
+            <div class="song-entity-analysis">
+              <Typography variant="h4">Entity Analysis</Typography>
+              <List>
+                {songInfo.entityAnalysis.map((wordInfo, index) => (
+                  <ListItem key={index + 1}>
+                    {wordInfo.word +
+                      " - " +
+                      wordInfo.salience +
+                      " - " +
+                      wordInfo.category}
```
amaurymedeiros, line 173: nit: probably it's better to use something like  [react-string-format)](https://www.npmjs.com/package/react-string-format) for that. It would look like `format('{0} - {1} - {2}', wordInfo.word, wordInfo.salience, wordInfo.category)`

amaurymedeiros (commented):
LGTM.

src/main/webapp/src/components/songInfo.js
jeinygroove: sure, it was made just to display info, because it'll change very soon by charts

jgaillard85 (approved):

PR #14, opened 2020-08-21
Add method to save to database

sashapff:


src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -2,15 +2,47 @@
...
 /** Class that provides methods to work with the database. */
 public class SongDataBase {
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private static final String ARTIST_PROPERTY = "artist";
+  private static final String NAME_PROPERTY = "name";
+  private static final String ALBUM_PROPERTY = "album";
+  private static final String COUNTER_PROPERTY = "counter";
+  private static final String SONG_PROPERTY = "Song";
+
+  /** Increase the counter of song's requests if it in the database, otherwise store a new song. */
```
jeinygroove: At this moment you don't increase counter, so it's better to change documentation for this method just for now.

jeinygroove (approved):

src/main/java/com/google/alpollo/Song.java
```diff
@@ -12,7 +12,7 @@ public Song(String artist, String name, String album) {
     this.album = album;
   }
 
-  /** Returns artists's name of the song. */
+  /** Returns artist's name of the song. */
```
amaurymedeiros: nit: maybe "Returns the name of the song's artists"?

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -2,15 +2,47 @@
...
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
 
 /** Class that provides methods to work with the database. */
 public class SongDataBase {
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private static final String ARTIST_PROPERTY = "artist";
+  private static final String NAME_PROPERTY = "name";
+  private static final String ALBUM_PROPERTY = "album";
+  private static final String COUNTER_PROPERTY = "counter";
```
amaurymedeiros: Add some documentation to the counter property, as it's not clear what it represents. Maybe choosing a name that's more explicit/obvious is enough. The other properties, for example, don't need documentation because what they represent is obvious.

sashapff: It is the number of times song was requested. I can add it to documentation because I don't know how to call it in another way.

amaurymedeiros: Maybe `SEARCH_COUNTER`? I think in any case it deserves a documentation, because it will be better to explain what it does. "Number of times users searched for this song in the system.", or something similar.

jgaillard85 (approved):

PR #17, opened 2020-08-24
fix getProjectID for running application

jeinygroove:
Situation: for tests we need config.json in `src/test/resources`, for running application, surprise, `src/main/resources` won't work, we need this file in `WEB_INF`, so I changed `getProjectID` function and now it tries to open both streams, because we don't know are we testing or running. 

This code doesn't look "DRY", but I haven't found how to make it more accurate. Please, suggest smth :(

P.S. maybe there's one folder that will work for both tests and run...


src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -35,11 +34,16 @@ private ConfigHelper() {
...
    */
   public static String getProjectID() {
-    try (InputStream inputStream = ClassLoader.getSystemResourceAsStream(CONFIG_FILE_NAME);
+    try (InputStream inputStream = new FileInputStream(new File(CONFIG_FILE_PATH));
          final Reader reader = new InputStreamReader(Objects.requireNonNull(inputStream))) {
       return gson.fromJson(reader, ConfigInfo.class).getProjectID();
     } catch (Exception parseException) {
-      return null;
+      try (InputStream inputStream = ClassLoader.getSystemResourceAsStream(CONFIG_FILE_NAME);
+           final Reader reader = new InputStreamReader(Objects.requireNonNull(inputStream))) {
```
sashapff: The same code as before. Better to make a function for this part. And what exceptions do you catch?

jeinygroove: yes, I don't know how don't repeat it and make it look normal

jeinygroove: I catch any exception actually, more specific nullpointer and json exceptions

RazvanBudacaTechquarter (approved):

src/main/java/com/google/alpollo/ConfigHelper.java
```diff
@@ -35,7 +35,11 @@ private ConfigHelper() {
    * or it was incorrect/didn't have projectID field
    */
   public static String getProjectID() {
-    try (InputStream inputStream = ClassLoader.getSystemResourceAsStream(CONFIG_FILE_NAME);
+    File configFile = new File(CONFIG_FILE_PATH);
+
+    try (InputStream inputStream = (configFile.exists()) ? 
+           gnew FileInputStream(configFile) :
```
jgaillard85: typo

jgaillard85 (commented):
What is the issue with having config.json in src/main/resources when running the app?

jeinygroove:
> What is the issue with having config.json in src/main/resources when running the app?

application doesn't see it((

[This is the reference](https://cloud.google.com/appengine/docs/standard/java/config/appref#Java_appengine_web_xml_Static_files_and_resource_files)

But I've not found a good explanation why `src/main/resources` won't work and why `CONFIG_FILE_PATH` won't work for test's

jgaillard85:
Have you tried using javax.servlet.ServletContext.getResourceAsStream instead of ClassLoader.getResourceAsStream?

jeinygroove:
> javax.servlet.ServletContext.getResourceAsStream

No... I'll try now, sounds genius.

jeinygroove:
Looks like I should create MockServletContext for testing then, ufff

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -15,4 +15,12 @@ public SongSentiment(float score, float magnitude) {
...
     this.magnitude = magnitude;
   }
+
+  public float getScore() {
+    return score;
+  }
+
+  public float getMagnitude() {
+    return magnitude;
+  }
```
jgaillard85, line 25: This doesn't seem to belong in this PR

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -28,31 +28,19 @@ public SongEntity(String name, double salience, String type, String wikiLink) {
...
   @Override
   public boolean equals(Object obj) {
     if (obj == this) {
       return true;
     }
 
-    if (!(obj instanceof SongEntity)) { 
-      return false; 
-  } 
+    if (!(obj instanceof SongEntity)) {
```
jgaillard85: Changes in this file don't seem to belong to this PR

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -34,14 +43,49 @@
...
+
+    entityServletUnderTest = new EntityServlet() {
+      @Override
+      public ServletContext getServletContext() {
+        return mockServletContext;
+      }
+    };
+
+    when(mockServletContext.getResourceAsStream(anyString())).thenAnswer(
+        (Answer<InputStream>) invocation -> new FileInputStream(CONFIG_FILE_PATH));
```
jgaillard85: Here you could be a bit less generic, and use the argument of getResourceAsStream in the answer, as what you want to change is not the entire thing, but just the root path. That way, if you ever need the ServletContext to access other resources, the mockServletContext will still work

src/main/java/com/google/alpollo/model/SongSentiment.java
jeinygroove: This change was made for new tests. Before it Budi just called `getSentiment` method and received `Sentiment` object, but now we have mock servlets and do post request and receive `SongSentiment` directly. So I need getters for it.

jgaillard85: Got it

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -34,14 +42,49 @@
...
+
+    entityServletUnderTest = new EntityServlet() {
+      @Override
+      public ServletContext getServletContext() {
+        return mockServletContext;
+      }
+    };
+
+    when(mockServletContext.getResourceAsStream(anyString()))
+      .thenReturn(new FileInputStream(CONFIG_FILE_PATH));
```
jgaillard85: Here I was thinking of returning something like
`new FileInputStream(TEST_RESOURCE_PATH + invocation.getArgument(0))`
with `TEST_RESOURCE_PATH = System.getProperty("user.dir") + "/src/main/webapp"`

That way, you would have the possibility of accessing more than just config.json if the need arises

jeinygroove: aaaaa I got it, makes sense

jgaillard85 (commented):
LGTM

PR #18, opened 2020-08-24
Find lyrics

jeinygroove:


src/main/webapp/src/components/songInfo.js
```diff
@@ -51,54 +52,53 @@ const styles = (theme) => ({
...
-        "Sailors fighting in the dance hall\n" +
-        "Oh man, look at those cavemen go\n" +
-        "It's the freakiest show\n" +
-        "Take a look at the lawman\n" +
-        "Beating up the wrong guy\n" +
-        "Oh man, wonder if he'll ever know\n" +
-        "He's in the best selling show\n" +
-        "Is there life on Mars?",
+      bandName: this.state.artistName.toUpperCase(),
+      songName: this.state.songName.toUpperCase(),
```
RazvanBudacaTechquarter, line 94: Why are we changing the names to upper case?

src/main/webapp/src/components/songInfo.js
```diff
@@ -54,26 +54,38 @@ class SongInfo extends React.Component {
...
+    } catch (_) {}
   }
 
   render() {
     const classes = this.props.classes;
 
     if (this.state == undefined) {
       return <div className={classes.root}>
-        <p>State wasn't defined, please use search page.</p>
+        <p>State wasn't defined, please, try use search page.</p>
```
RazvanBudacaTechquarter: I don't think this is written well. It should be something like
State wasn't defined, please try to use the search page.

But does the user know what "state" means?

src/main/webapp/src/components/songInfo.js
jeinygroove: Because this fields are given by user on the search page, so if he will write "YeStErDaY" api will find the lyrics, but it's not pretty to show string like this on the page))

RazvanBudacaTechquarter (approved):

src/main/webapp/src/components/search.js
```diff
@@ -127,6 +164,20 @@ class Search extends React.Component {
...
+          if (this.state.error)
+            return (
+              <div>
+                <p>{this.state.error.message}</p>
+              </div>
+            );
+          if (this.state.isLoading)
+            return (
+              <div>
+                <CircularProgress style={{ color: "black" }} />;
```
amaurymedeiros: The trailing `;` is displayed next to the loading animation

image

amaurymedeiros (approved):
Only a minor change

PR #19, opened 2020-08-24
Add more info to our entities

RazvanBudacaTechquarter:
Now our entities have additional metadata (like wikipedia links) and a special type (like Person, Work of Art etc.).


RazvanBudacaTechquarter:
The new test seems to be flaky, mostly because of the order of elements in the API generated Map.

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -9,12 +11,23 @@
  */
 public class SongEntity {
   private final String name;
-  // Each entity has a salience score, telling us how important the word is in the given context.
+  /**  Each entity has a salience score, telling us how important the word is in the given context. */
   private final double salience;
+  /** 
+   * The API sets each entity to a category.
+   * e.g. PERSON, LOCATION, WORK_OF_ART*/
```
jeinygroove: move `*/` to the next line

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -9,12 +11,23 @@
...
+  /**  Each entity has a salience score, telling us how important the word is in the given context. */
   private final double salience;
+  /** 
+   * The API sets each entity to a category.
+   * e.g. PERSON, LOCATION, WORK_OF_ART*/ 
+  private final String type;
+  /** 
+   * For most entity types, the metadata is a Wikipedia URL. Other times, it can be extra info.
+   * An object containing a list of "key": value pairs. Example: { "name": "wrench", "mass": "1.3kg", "count": "3" }.
+  */
```
jeinygroove: nit: add space to the left of `*/`

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -9,12 +11,23 @@
...
-  // Each entity has a salience score, telling us how important the word is in the given context.
+  /**  Each entity has a salience score, telling us how important the word is in the given context. */
   private final double salience;
+  /** 
+   * The API sets each entity to a category.
+   * e.g. PERSON, LOCATION, WORK_OF_ART*/ 
+  private final String type;
+  /** 
+   * For most entity types, the metadata is a Wikipedia URL. Other times, it can be extra info.
+   * An object containing a list of "key": value pairs. Example: { "name": "wrench", "mass": "1.3kg", "count": "3" }.
```
jeinygroove: `"key": "value"` or `key-value`, or smth like this

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -9,12 +11,23 @@
...
+  */
+  private final Map<String, String> metadata;
```
jeinygroove: I don't think that we need all metadata, because we have only an idea of how to use `Wikipedia URL` now and not smth else.

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -27,7 +27,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       String lyrics = request.getParameter(LYRICS);
 
       // Get the Entity list from the API
-      List<Entity> entityList = new ArrayList<>(AnalysisHelper.getEntityList(lyrics));
+      List<Entity> entityList = AnalysisHelper.getEntityList(lyrics);
```
jeinygroove, line 29: why you changed this and does it connected with this pr?

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -60,13 +62,15 @@ public void checkSentimentMagnitude() throws IOException {
    */
   @Test
   public void top10SalientEntitiesWithLessThan10Entities() throws IOException {
-    List<Entity> entityList = new ArrayList<>(AnalysisHelper.getEntityList(LYRICS_SHORT));
+    List<Entity> entityList = AnalysisHelper.getEntityList(LYRICS_SHORT);
```
jeinygroove, line 65: why you changed this and does it connected with this pr?

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -76,15 +80,38 @@ public void top10SalientEntitiesWithLessThan10Entities() throws IOException {
    */
   @Test
   public void top10SalientEntitiesWithMoreThan10Entities() throws IOException {
-    List<Entity> entityList = new ArrayList<>(AnalysisHelper.getEntityList(LYRICS_LONG));
+    List<Entity> entityList = AnalysisHelper.getEntityList(LYRICS_LONG);
```
jeinygroove, line 83: why you changed this and does it connected with this pr?

src/main/java/com/google/alpollo/EntityServlet.java
RazvanBudacaTechquarter: All of these were changed in a previous PR, I don't know how they slipped past me. It's not necessary to make a copy of the list.

src/main/java/com/google/alpollo/model/SongEntity.java
RazvanBudacaTechquarter: I'm not sure we can remove stuff from the Map, I'll research a bit.

src/main/java/com/google/alpollo/EntityServlet.java
jeinygroove: You need to find a branch in which they're made and upload changes from it to this one

src/main/java/com/google/alpollo/model/SongEntity.java
jeinygroove: you just get map with metadata using `entity.getMetadataMap()` and get the value of the key `wikipedia_url` (or how it's called) and that's all, if this key doesn't exist, then you can set it to null or smth like that, as you will decide

src/main/java/com/google/alpollo/EntityServlet.java
RazvanBudacaTechquarter: They aren't changed on that branch either, maybe I just forgot or something. Isn't it better just to change it real fast from this branch?

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -76,15 +80,38 @@ public void top10SalientEntitiesWithLessThan10Entities() throws IOException {
...
+    List<Entity> entityList = AnalysisHelper.getEntityList(LYRICS_WITH_METADATA);
+    List<SongEntity> simplifiedEntityList = AnalysisHelper.getSimplifiedEntityList(entityList);
+
+    Map<String, String> metadata = new HashMap<>();
+    metadata.put("mid", "/m/045c7b");
+    metadata.put("wikipedia_url", "https://en.wikipedia.org/wiki/Google");
+
+    String actual = gson.toJson(AnalysisHelper.getTopSalientEntities(simplifiedEntityList));
+    String expected = gson.toJson(Arrays.asList(new SongEntity("Google", 0.88, "ORGANIZATION", metadata),
+        new SongEntity("results", 0.12, "OTHER", Collections.emptyMap())));
```
jeinygroove: I think the best way to compare `SongEntity` objects it to override `equals`, which will compare metadata as sets (unordered), so if in the future we'll add not only Wikipedia URL, but smth else, we won't need to change these tests.  And compare objects instead of json strings

RazvanBudacaTechquarter: I tried comparing objects and it did not really work, it always fails. Maybe I don't know how?

jeinygroove: how did you compare them?

src/main/java/com/google/alpollo/EntityServlet.java
jeinygroove: I think you can leave it as it is then

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: `Assert.assertEquals(list, list)`
All the objects had different IDs
I'm trying now with `Assert.assertThat()`.

RazvanBudacaTechquarter: Ok your idea seems the best, I could override `equals` and compare the objects in 2 for loops, to avoid converting them to strings.
Maybe the guys have a better idea? Let's check with them too.

jeinygroove: Also you can work not with lists of `Entities`, but with sets or another unordered collection, if the order doesn't matter, and in this case there's no need to do loops

jeinygroove (approved):

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: I think order matters in our case, we're talking about top 10 most important entities.

jeinygroove: so, I think that overriding `equals` and `assertThat(actual, is(expected));` should work

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +35,26 @@ public int compare(SongEntity a, SongEntity b) {
...
+      return true;
+    }
+
+    if (!(obj instanceof SongEntity)) { 
+      return false; 
+  } 
+
+    SongEntity entity = (SongEntity) obj;
+    return salience == entity.salience
+        && (name == entity.name || (name != null && name.equals(entity.getName())))
```
jeinygroove: you can just replace it with `Objects.equals(name, entity.name)`. Why you use `entity.name` and then `entity.getName()`?

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +35,26 @@ public int compare(SongEntity a, SongEntity b) {
...
+        && (type == entity.type || (type != null && type .equals(entity.getType())))
```
jeinygroove: Same. Also you have a space between `type` and `.equals..`

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +35,26 @@ public int compare(SongEntity a, SongEntity b) {
...
+        && (wikiLink == entity.wikiLink || (wikiLink != null && wikiLink .equals(entity.getWikiLink())));
```
jeinygroove: Same. And same thing about space.

src/main/java/com/google/alpollo/model/SongEntity.java
RazvanBudacaTechquarter: That's how I've seen it done by more people on the internet and I thought:
"Hey, maybe it's a good practice, let's see what the others have to say about it".

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +35,21 @@ public int compare(SongEntity a, SongEntity b) {
...
+      return true;
+    }
+
+    if (!(obj instanceof SongEntity)) { 
+      return false; 
+  } 
+
+    SongEntity entity = (SongEntity) obj;
+    return salience == entity.salience
+        && (name.equals(entity.name))
```
jeinygroove: `Objects.equals(name, entity.name)`, because it will fail with exception if `name == null`

jeinygroove (approved):
After this fix, LGTM

src/main/java/com/google/alpollo/model/SongEntity.java
RazvanBudacaTechquarter: But name can't be null, it will always have a value. That's why I chose string.equals().

jeinygroove: Logically can't, but you don't stop me from doing this: `SongEntity newSongEntity = new SongEntity(null, 0.0, null, null);`

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -10,6 +10,7 @@
 import com.google.cloud.language.v1.Document;
 import com.google.cloud.language.v1.EncodingType;
 import com.google.cloud.language.v1.Entity;
+import com.google.cloud.language.v1.EntityMention;
```
amaurymedeiros: nit: You added a new import, but doesn't seem to be using it... Clean the unused imports

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +35,21 @@ public int compare(SongEntity a, SongEntity b) {
       return Double.compare(b.salience, a.salience);
     }
   };
+
+  @Override
+  public boolean equals(Object obj) {
```
amaurymedeiros, line 42: When you implement `equals`, you should also implement `hashCode`

https://www.baeldung.com/java-equals-hashcode-contracts

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -9,12 +9,21 @@
...
 public class SongEntity {
   private final String name;
-  // Each entity has a salience score, telling us how important the word is in the given context.
+  /**  Each entity has a salience score, telling us how important the word is in the given context. */
   private final double salience;
+  /** 
+   * The API sets each entity to a category.
+   * e.g. PERSON, LOCATION, WORK_OF_ART
+   */ 
+  private final String type;
```
amaurymedeiros: How do we get this string (i.e. "PERSON", "LOCATION"), etc? Is it straight from the API? If it is, we need to make it explicit in the docs, i.e. say it matches the entities from _____ API and preferably include a link pointing to all entities available.

If instead you know in advance all the possible types and are setting them manually, maybe we should use an `enum` here instead of a string.

src/main/java/com/google/alpollo/model/SongEntity.java
amaurymedeiros: Olya has a point here. You shouldn't trust your assumptions when designing APIs. In larger projects, we use annotations when we want to make sure a field is not null. For example, the constructor would be something like `public SongEntity(@NotNull String name, double salience, @Nullable String type, String wikiLink)`. That's not the case here, so we need to be prepared for someone passing `null`.

Also, in `wikiLink` documentation, you say that entities **might**  have one, so at least this field can be null, right?

RazvanBudacaTechquarter: I made sure to set `wikiLink` to an empty string if it's ever null, but I got your point, `Object.equals()` just covers a wider range of cases and it's better for project safety.

src/main/java/com/google/alpollo/model/SongEntity.java
RazvanBudacaTechquarter: Awesome, lots of new things!

jeinygroove: Totally forgot to mention this thing when told about `equals()`

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +37,31 @@ public int compare(SongEntity a, SongEntity b) {
...
+
+    if (!(obj instanceof SongEntity)) { 
+      return false; 
+  } 
+
+    SongEntity entity = (SongEntity) obj;
+    return salience == entity.salience
+        && (Objects.equal(name,entity.name))
+        && (Objects.equal(type, entity.type))
+        && (Objects.equal(wikiLink,entity.wikiLink));
```
amaurymedeiros: nit: add whitespace after `,`

amaurymedeiros (commented):
LGTM

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -26,4 +37,31 @@ public int compare(SongEntity a, SongEntity b) {
...
+  }
+
+  @Override
+  public int hashCode() {
+    int hash = 7;
+    hash = 31 * hash + (int) salience;
+    hash = 31 * hash + (name == null ? 0 : name.hashCode());
+    hash = 31 * hash + (type == null ? 0 : type.hashCode());
+    hash = 31 * hash + (wikiLink == null ? 0 : wikiLink.hashCode());
```
jgaillard85: If you want to be safer here, you can use Objects.hashcode() that will handle potential null values

jgaillard85 (approved):

PR #21, opened 2020-08-25
Budi added a cool chart

RazvanBudacaTechquarter:
Pressing the button will call the API to load and draw a chart with the most important words of the text.

For now it's all static data. To convert, we just call fetch to our server and use the list from the response to get info about our entities.

### DIFF
--- a/src/main/webapp/src/components/songInfo.js
+++ b/src/main/webapp/src/components/songInfo.js
@@ -189,17 +189,8 @@ class SongInfo extends React.Component {
             </div>
             <div class="song-entity-analysis">
               <Typography variant="h4">Entity Analysis</Typography>
-              <List>
-                {songInfo.entityAnalysis.map((wordInfo, index) => (
-                  <ListItem key={index + 1}>
-                    {wordInfo.word +
-                      " - " +
-                      wordInfo.salience +
-                      " - " +
-                      wordInfo.category}
-                  </ListItem>
-                ))}
-              </List>
+              <button onClick={loadChartAPI}>Submit</button>
+              <div id="chart-container"></div>
             </div>
           </div>
         </div>
--- a/src/main/webapp/src/components/songInfo.js
+++ b/src/main/webapp/src/components/songInfo.js
@@ -106,32 +106,32 @@ class SongInfo extends React.Component {
       entityAnalysis: [
         {
           word: "affair",
-          salience: "0.51",
+          salience: 0.51,
           category: "work of art",
         },
         {
           word: "girl",
-          salience: "0.09",
+          salience: 0.09,
           category: "person",
         },
         {
           word: "film",
-          salience: "0.06",
+          salience: 0.06,
           category: "work of art",
         },
         {
           word: "selling show",
-          salience: "0.04",
+          salience: 0.04,
           category: "work of art",
         },
         {
           word: "God",
-          salience: "0.01",
+          salience: 0.01,
           category: "person",
         },
         {
           word: "Mars",
-          salience: "0.01",
+          salience: 0.01,
           category: "location",
         },
       ],
--- a/src/main/webapp/src/components/songInfo.js
+++ b/src/main/webapp/src/components/songInfo.js
@@ -191,8 +191,7 @@ class SongInfo extends React.Component {
             </div>
             <div class="song-entity-analysis">
               <Typography variant="h4">Entity Analysis</Typography>
-              <button onClick={loadChartAPI}>Get Entity Analysis results</button>
-              <div id="chart-container"></div>
+              <div id="chart-container" ref={() => loadChartAPI()}></div>
             </div>
           </div>
         </div>
--- a/src/main/webapp/src/components/songInfo.js
+++ b/src/main/webapp/src/components/songInfo.js
@@ -146,7 +146,7 @@ class SongInfo extends React.Component {
       ],
     };
 
-    function drawChart() {
+    function drawChart(element) {
       const data = new google.visualization.DataTable();
       data.addColumn('string', 'Name');
       data.addColumn('number', 'Salience');
@@ -160,19 +160,18 @@ class SongInfo extends React.Component {
         height: 500
       };
 
-      const entityChart = new google.visualization.PieChart(document.getElementById('chart-container'));
+      const entityChart = new google.visualization.PieChart(element);
       entityChart.draw(data, options);
     }
 
-    function loadChartAPI() {
+    function loadChartAPI(element) {
       // Load the Visualization API and the corechart package.
       google.charts.load('current', {'packages':['corechart']});
 
       // Set a callback to run when the Google Visualization API is loaded.
-      google.charts.setOnLoadCallback(drawChart);
+      google.charts.setOnLoadCallback(function() { drawChart(element) });
     }
 
-    // TODO display entities with charts
     return (
       <div className={classes.root}>
         <Typography variant="h3" className={classes.songName}>
@@ -191,7 +190,7 @@ class SongInfo extends React.Component {
             </div>
             <div class="song-entity-analysis">
               <Typography variant="h4">Entity Analysis</Typography>
-              <div id="chart-container" ref={() => loadChartAPI()}></div>
+              <div ref={(elem) => { if (elem) loadChartAPI(elem); }}></div>
             </div>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


src/main/webapp/src/components/songInfo.js
```diff
@@ -144,6 +146,32 @@ class SongInfo extends React.Component {
...
+    function drawChart() {
+      const data = new google.visualization.DataTable();
+      data.addColumn('string', 'Name');
+      data.addColumn('number', 'Salience');
+      songInfo.entityAnalysis.forEach((entity) => {
+        data.addRow([entity.word, entity.salience]);
+      });
+
+      const options = {
+        'title': 'Most important words',
```
jeinygroove: you can remove `'` around `title`

src/main/webapp/src/components/songInfo.js
```diff
@@ -144,6 +146,32 @@ class SongInfo extends React.Component {
...
+        'width':600,
```
jeinygroove: You can remove `'` around `width`. And add space after `:`

src/main/webapp/src/components/songInfo.js
```diff
@@ -144,6 +146,32 @@ class SongInfo extends React.Component {
...
+        'height':500
```
jeinygroove: You can remove `'` around `height`. And add space after `:`

src/main/webapp/src/components/songInfo.js
```diff
@@ -163,17 +191,8 @@ class SongInfo extends React.Component {
...
-                  <ListItem key={index + 1}>
-                    {wordInfo.word +
-                      " - " +
-                      wordInfo.salience +
-                      " - " +
-                      wordInfo.category}
-                  </ListItem>
-                ))}
-              </List>
+              <button onClick={loadChartAPI}>Get Entity Analysis results</button>
```
jeinygroove: Of course we don't want force user to push some button just to see the charts. One of the best solution is to use `ref`.

src/main/webapp/src/components/songInfo.js
```diff
@@ -163,17 +191,8 @@ class SongInfo extends React.Component {
...
+              <div id="chart-container"></div>
```
jeinygroove: Add `ref={() => loadChartAPI()}` and remove button that you had, so this function will be called after rendering.

More info [here](https://reactjs.org/docs/refs-and-the-dom.html)

src/main/webapp/src/components/songInfo.js
RazvanBudacaTechquarter: Oh ok I thought we wanted a button. My bad.

jeinygroove (commented):
I also found a better solution with `ref`. In my first suggestion you still get an element with `getElementById`, but `ref` provides you an opportunity to get this element immediately. 

For this you need to add `ref={(elem) => {
                  if (elem) loadChartAPI(elem);
                }}`
Short explanation: we need to check if `elem` is not null, because ref is called twice, one time with null and one time with loaded element.

Then you need to add `element` argument to `loadChartAPI` and `drawChart` functions.

Then you need to replace `google.charts.setOnLoadCallback(drawChart);` with `google.charts.setOnLoadCallback(function() { drawChart(element) });`.  (You can't replace `function() { drawChart(element) }` with `drawChart(element)`, it's quite long explanation, I can tell it if you'll ask or I'll add it to this comment a little bit later).

And the main part: `const entityChart = new google.visualization.PieChart(element);` instead of calling `getElementById`

src/main/webapp/src/components/songInfo.js
```diff
@@ -144,6 +146,32 @@ class SongInfo extends React.Component {
...
+    function drawChart() {
+      const data = new google.visualization.DataTable();
+      data.addColumn('string', 'Name');
+      data.addColumn('number', 'Salience');
+      songInfo.entityAnalysis.forEach((entity) => {
+        data.addRow([entity.word, entity.salience]);
+      });
+
+      const options = {
+        title : 'Most important words',
```
jeinygroove: remove space between `title` and `:`. Same for other fields.

RazvanBudacaTechquarter: Spaces are my enemy.

RazvanBudacaTechquarter:
> I also found a better solution with `ref`. In my first suggestion you still get an element with `getElementById`, but `ref` provides you an opportunity to get this element immediately.
> 
> For this you need to add `ref={(elem) => { if (elem) loadChartAPI(elem); }}`
> Short explanation: we need to check if `elem` is not null, because ref is called twice, one time with null and one time with loaded element.
> 
> Then you need to add `element` argument to `loadChartAPI` and `drawChart` functions.
> 
> Then you need to replace `google.charts.setOnLoadCallback(drawChart);` with `google.charts.setOnLoadCallback(function() { drawChart(element) });`. (You can't replace `function() { drawChart(element) }` with `drawChart(element)`, it's quite long explanation, I can tell it if you'll ask or I'll add it to this comment a little bit later).
> 
> And the main part: `const entityChart = new google.visualization.PieChart(element);` instead of calling `getElementById`

I didn't quite get why this method is better in our case. Can you tell me again?

jeinygroove:
> > I also found a better solution with `ref`. In my first suggestion you still get an element with `getElementById`, but `ref` provides you an opportunity to get this element immediately.
> > For this you need to add `ref={(elem) => { if (elem) loadChartAPI(elem); }}`
> > Short explanation: we need to check if `elem` is not null, because ref is called twice, one time with null and one time with loaded element.
> > Then you need to add `element` argument to `loadChartAPI` and `drawChart` functions.
> > Then you need to replace `google.charts.setOnLoadCallback(drawChart);` with `google.charts.setOnLoadCallback(function() { drawChart(element) });`. (You can't replace `function() { drawChart(element) }` with `drawChart(element)`, it's quite long explanation, I can tell it if you'll ask or I'll add it to this comment a little bit later).
> > And the main part: `const entityChart = new google.visualization.PieChart(element);` instead of calling `getElementById`
> 
> I didn't quite get why this method is better in our case. Can you tell me again?

Because in this case you don't need to add `id` to the element and you don't need to get this element manually when you already "have" it. I don't want to say that this is better, it's a **little bit** better  because you won't depend on `id` value.

This also shows `ref`'s logic, we don't just use it to call a function after loading of the element, but actually **refer** to this element.

RazvanBudacaTechquarter:
> > > I also found a better solution with `ref`. In my first suggestion you still get an element with `getElementById`, but `ref` provides you an opportunity to get this element immediately.
> > > For this you need to add `ref={(elem) => { if (elem) loadChartAPI(elem); }}`
> > > Short explanation: we need to check if `elem` is not null, because ref is called twice, one time with null and one time with loaded element.
> > > Then you need to add `element` argument to `loadChartAPI` and `drawChart` functions.
> > > Then you need to replace `google.charts.setOnLoadCallback(drawChart);` with `google.charts.setOnLoadCallback(function() { drawChart(element) });`. (You can't replace `function() { drawChart(element) }` with `drawChart(element)`, it's quite long explanation, I can tell it if you'll ask or I'll add it to this comment a little bit later).
> > > And the main part: `const entityChart = new google.visualization.PieChart(element);` instead of calling `getElementById`
> > 
> > 
> > I didn't quite get why this method is better in our case. Can you tell me again?
> 
> Because in this case you don't need to add `id` to the element and you don't need to get this element manually when you already "have" it. I don't want to say that this is better, it's a **little bit** better because you won't depend on `id` value.
> 
> This also shows `ref`'s logic, we don't just use it to call a function after loading of the element, but actually **refer** to this element.

Sounds interesting, imma go for it.

jeinygroove (approved):

src/main/webapp/src/components/songInfo.js
```diff
@@ -144,7 +146,32 @@ class SongInfo extends React.Component {
       ],
     };
 
-    // TODO display entities with charts
+    function drawChart(element) {
```
amaurymedeiros: `drawChart` is a generic name. Please use a more meaningful one and/or add some docs to explain which chart we're drawing here (probably an entity pie chart) and giving details about how it's expected to look like

RazvanBudacaTechquarter:
Thank you Olya, very cool!

amaurymedeiros (commented):
LGTM

PR #23, opened 2020-08-25
Lyrics component

jeinygroove:
Just moved lyrics to the separate component.


sashapff (approved):

src/main/webapp/src/components/lyrics.js
```diff
@@ -0,0 +1,34 @@
...
+     * Lyrics, returned by lyrics.ovh has `\n\n`
+     * and `\r\n`.
+     * We replace this with one `\n` to
+     * make it look better.
+     */
+    const lyrics = this.props.lyrics
+      .split("\r\n")
+      .join("\n")
+      .split("\n\n")
+      .join("\n");
```
amaurymedeiros: Instead of converting the string back and forth into an array, you could use `replaceAll`.

jeinygroove: I used `replace` instead of `replaceAll`, because:

```
As of August 2020 the replaceAll() method is supported by Firefox but not by Chrome. It will become available in Chrome 85.
```

amaurymedeiros: Whaaaaat?! I didn't know that, sorry. I saw the banner in the mozilla documentation page, but it doesn't have a link anywhere, so I couldn't find more details. Have you found something?

Regarding the usage of `replace`, you need to use a regular expression, otherwise it will only replace the first occurrence. 

The good news is you only need one call, since `(\r\n|\n\n)` will match both sequence of characters you want to replace with `\n`.

jeinygroove: Yes, but I noticed a mistake, please have a look at new documentation

jeinygroove: I should have change `\n\n\n` with `\n\n`

jeinygroove: I saw this message about `replaceAll` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll)

amaurymedeiros: Oh, OK. But you still need to use a regex. Consider the following string:

`Here is my string\n\n\n. I wan't to replace some sequences\r\n. However, by using a string and not a regex\n\n\n, I don't replace all of them.\r\n"`

If you use a string as a parameter, you'll end up with the following string:
`Here is my string\n\n. I wan't to replace some sequences\n. However, by using a string and not a regex\n\n\n, I don't replace all of them.\r\n"`

If you use a regex instead, you'll achieve what you want:

`Here is my string\n\n. I wan't to replace some sequences\n. However, by using a string and not a regex\n\n, I don't replace all of them.\n"`

Let's discuss that in today's sync. I'm curious to understand more about the problem and clarify some things about line breaks.

----

Yea, I saw this banner, but there is no additional explanation. They just dropped the news and ran away.

jeinygroove: Yes, I already changed it to regexp

amaurymedeiros (commented):
LGTM

jgaillard85 (approved):

PR #25, opened 2020-08-26
Fetch sentiment analysis info and display it

jeinygroove:


sashapff (approved):

RazvanBudacaTechquarter (approved):
Maybe we should make the info icons a bit smaller.

jeinygroove:
> Maybe we should make the info icons a bit smaller.

Yes, thought about it, will make it a little bit late.

src/main/webapp/src/components/sentimentAnalysisInfo.js
```diff
@@ -0,0 +1,121 @@
...
+        <div>
+          <p>{this.state.error.message}</p>
+        </div>
+      );
+    }
+
+    if (this.state.isLoading) {
+      return (
+        <div>
+          <CircularProgress style={{ color: "black" }} />
```
amaurymedeiros, line 87: nit: can we centralize the style of the CircularProgress somewhere? We should be consistent across all the parts of the app, so there is no need to set `"color: black"` everywhere, because if we change our mind we need to replace it everywhere.

This change should not be part of this Pull Request, since it's unrelated.

amaurymedeiros (commented):
LGTM

src/main/webapp/src/components/sentimentAnalysisInfo.js
jeinygroove: sure

jgaillard85 (approved):

PR #26, opened 2020-08-27
Add counter to search requests of the song

sashapff:


jeinygroove (commented):
Can you add tests for this?

sashapff:
> Can you add tests for this?

I will add tests in another PR. Let's merge it to objectify branch for now))

jeinygroove:
> > Can you add tests for this?
> 
> I will add tests in another PR. Let's merge it to objectify branch for now))

Ok

jeinygroove (approved):

src/main/java/com/google/alpollo/SongCounter.java
```diff
@@ -0,0 +1,22 @@
+package com.google.alpollo;
+
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Index;
+
+/** Represents the structure of the song to save search counter in the database. */
+@Entity
+public class SongCounter extends Song {
```
jgaillard85: Why not add searchCounter as a field to the Song class instead of creating this new class?
This inheritance also does not make sense from an Object Oriented point of view. SongCounter is not a particular sub category of Song.

src/main/java/com/google/alpollo/SongCounter.java
```diff
@@ -0,0 +1,31 @@
...
+/** Represents the structure of the song to save search counter in the database. */
+@Entity
+public class SongCounter {
+  @Id private String id;
+  @Index private Song song;
+  /** Number of times users searched for this song in the system. */
+  @Index private int searchCounter = 0;
+
+  /** Objectify requires no argument constructor. Do not use it. */
+  protected SongCounter() { }
```
jeinygroove: Why it's `protected` if you tell in the comment not to use it? Can it be used by derived classes?

src/main/java/com/google/alpollo/Song.java
```diff
@@ -2,35 +2,26 @@
...
 import com.googlecode.objectify.annotation.Entity;
-import com.googlecode.objectify.annotation.Id;
 
 /** 
  * Represents the structure of the song. 
  * Please do not add fields because they will be saved in the search counter database. 
  */
 @Entity
 public class Song {
-  @Id private String id;
```
jeinygroove, line 13: Same question: why not to store this id and get rid of song counter class?

sashapff: Because I don't want to save counter as a part of song

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -1,19 +1,28 @@
...
+    OfyService.ofy().save().entity(songCounter).now();
   }
 
-  /** Returns the list of at most 10 songs. */
+  /** Returns the list of the most requested songs. */
   public static List<Song> topSongs() {
-    return OfyService.ofy().load().type(Song.class).limit(TOP_SIZE).list();
+    List<Song> songs = new ArrayList<>();
+    OfyService.ofy().load().type(SongCounter.class).order("-searchCounter").limit(TOP_SIZE).list()
+        .forEach(songCounter -> songs.add(songCounter.getSong()));
```
jeinygroove: Now `Song` class doesn't contain id field, which we need to know on the client side for retrieving `SongInfo`, can you, please, return `SongCounter`s instead: 
`return OfyService.ofy().load().type(SongCounter.class).order("-searchCounter").limit(TOP_SIZE).list();`

src/main/java/com/google/alpollo/OfyService.java
```diff
@@ -8,7 +8,7 @@
 public class OfyService {
   /** Registers Song class in Objectify. */
```
amaurymedeiros: update the documentation (it should say `SongCounter` now)

src/main/java/com/google/alpollo/Song.java
```diff
@@ -45,4 +33,10 @@ public String getName() {
   public String getAlbum() {
     return album;
   }
+
+  /** Calculates and returns song id by each song. */
+  static String id(Song song) {
+    // TODO: add hashcode instead of concatenation
+    return song.artist + song.name + song.album;
+  }
```
amaurymedeiros: Why is this a static method that receives a song instead of an instance method with no parameters that returns `this.artist + this.name + this.album`?

src/main/java/com/google/alpollo/SongCounter.java
```diff
@@ -0,0 +1,31 @@
...
+    this.song = song;
+    id = Song.id(song);
+  }
+
+  Song getSong() {
+    return song;
+  }
+
+  /** Increases by 1 search counter. */
+  void increaseSearchCounter() {
```
amaurymedeiros: nit: maybe it's better to use  "increment" instead of "increase" in this case

jeinygroove (approved):

src/main/java/com/google/alpollo/Song.java
```diff
@@ -41,8 +29,8 @@ public String getName() {
     return name;
   }
 
-  /** Returns the name of the song's album. */
-  public String getAlbum() {
-    return album;
+  /** Calculates and returns song id by each song. */
+  protected Long id() {
```
amaurymedeiros: Why are we using the `protected` visibility here? Should this method be public?

src/main/java/com/google/alpollo/SongCounter.java
```diff
@@ -0,0 +1,31 @@
...
+  public SongCounter(Song song) {
+    this.song = song;
+    id = song.id();
+  }
+
+  Song getSong() {
+    return song;
+  }
+
+  /** Increments by 1 search counter. */
```
amaurymedeiros: Nit: "Increments search counter by 1"

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -1,19 +1,25 @@
...
 import java.util.List;
+import java.util.ArrayList;
 
 /** Class that provides methods to work with the database. */
 public class SongDataBase {
   /** Number of the songs that will be shown to the user. */
   private static final int TOP_SIZE = 10;
 
-  /** Save request song to database. */
+  /** Save request song to database and increase the search counter. */
```
amaurymedeiros: nit: increase -> increment

amaurymedeiros (approved):

PR #27, opened 2020-08-27
Entity analysis section

jeinygroove:
@BudiGabe It wasn't a good idea to write Charts as we did in portfolio projects, because JS implementation of Charts API manipulates with DOM elements directly. It's bad, because in React you shouldn't do it, it badly affects life-cycle of the application. 

As the result, we can receive this exception: 
`DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

I didn't knew that and just noticed it now, that's why I haven't insisted to use `react-google-chart` instead.

### DIFF
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -73,7 +73,7 @@ class EntityAnalysisInfo extends React.Component {
     if (this.state.isLoading) {
       return (
         <div>
-          <CircularProgress style={{ color: "black" }} />;
+          <CircularProgress style={{ color: "black" }} />
         </div>
       );
     }
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -1,0 +1,128 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { withStyles } from "@material-ui/styles";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import axios from "axios";
+
+const google = window.google;
+
+const styles = (theme) => ({
+  root: {
+    display: "flex",
+    flexDirection: "column",
+  },
+});
+
+/**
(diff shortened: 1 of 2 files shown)


jeinygroove:
I will rewrite charts to `Chart` component in this pr soon.

RazvanBudacaTechquarter:
> I will rewrite charts to `Chart` component in this pr soon.

Hmm I didn't expect that. But it shouldn't take too much to rewrite it.

jeinygroove:
Changed my mind, I will do an issue for this and somebody can fix it later.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    if (this.state.entityAnalysisInfo.length == 0) {
+      return <p>No entities was found.</p>;
+    }
+
+    /**
+     * Function will draw a pie chart with static data representing the most important words in
+     * our song's context.
+     * TODO fetch data from EntityServlet.
```
RazvanBudacaTechquarter: You can remove the TODO now.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -0,0 +1,132 @@
...
+    if (this.state.isLoading) {
+      return (
+        <div>
+          <CircularProgress style={{ color: "black" }} />;
+        </div>
+      );
+    }
+
+    if (this.state.entityAnalysisInfo.length == 0) {
+      return <p>No entities was found.</p>;
```
RazvanBudacaTechquarter: This should be rephrased.
Maybe
`No entities were found` or just `No entities found`

RazvanBudacaTechquarter (commented):
Seems very similar to the sentiment one.

RazvanBudacaTechquarter (approved):

amaurymedeiros (commented):
LGTM

jgaillard85 (approved):

PR #31, opened 2020-08-27
Display rating list

jeinygroove:


src/main/webapp/src/components/ratingList.js
```diff
@@ -31,71 +33,74 @@ function EnumeratedList(props) {
...
+  }
+
   render() {
     const classes = this.props.classes;
+    const topSongs = this.state.topSongs;
+
+    if (this.state.error) {
+      return (
+        <div>
+          <p>{this.state.error.message}</p>
```
RazvanBudacaTechquarter: Can you give an example of error? I'm thinking that maybe the user will have no idea what it actually says

RazvanBudacaTechquarter (changes requested):
All good besides the shady error message.

src/main/webapp/src/components/ratingList.js
jeinygroove: It depends on the servlet actually. Usually, it displays error that the servlet can send.

RazvanBudacaTechquarter: Maybe we could print this error to console so we know what's happening, but show to the user only something like
`Oops, something went wrong, please try again`

RazvanBudacaTechquarter (approved):

sashapff (approved):

src/main/webapp/src/components/ratingList.js
```diff
@@ -31,71 +33,75 @@ function EnumeratedList(props) {
...
+        })
+      );
+  }
+
   render() {
     const classes = this.props.classes;
+    const topSongs = this.state.topSongs;
+
+    if (this.state.error) {
+      console.log(this.state.error.message);
```
jgaillard85, line 80: In the future, it might good to think of a proper way to log messages on the server side

jgaillard85 (approved):

PR #33, opened 2020-08-28
Budi youtube

RazvanBudacaTechquarter:
For now, the code searches on YouTube by the most important word and prints in console the first 5 search results.

Displaying the videos in the client has been passed to Olya for faster development.


jeinygroove (approved):

jeinygroove:
Why you used `gapi`, have you seen similar code somewhere or there was some tutorial with it?

RazvanBudacaTechquarter:
> Why you used `gapi`, have you seen similar code somewhere or there was some tutorial with it?

Yep, [this](https://developers.google.com/youtube/v3/docs/search/list?apix=true) is what I used.

jeinygroove:
Anyway, this code works fine, but it **can't** be used, because you use sensitive info in it. For this task you should have done one of the next things: 

- Use Node JS server and retrieve keys from `.env` file or smth else (Cool thing, imho)
- Use Java code from the same API and add all these keys to `config.json`, which already was implemented (The easies solution I suppose, unfortunately I didn't knew about it when I was doing Node JS server).
- Put restrictions to the API key and Client ID (imho, not a good solution)
- Smth else...

Please, keep such things in mind for the future, so you and your team won't need to do extra work. Just remember that ALL things on client side are visible, even if they're hide somehow, there still be a way to retrieve them. Btw, I really liked [this](https://www.quora.com/Is-it-possible-to-protect-or-hide-the-client-side-JavaScript-code) Quora question answers about this topic.

PR #36, opened 2020-08-30
Node server to secure api keys

jeinygroove:
I used to get rid of `gapi` that @BudiGabe used in #33, because this code now should be on Node JS server and use `gapi` dependency instead of imported script, and it's bad because of the next reasons:

1. It's unmaintained [project](https://github.com/phated/node-gapi#readme) with the last commit dated by 2012.
2. It depends on `CoffeeScript`. And it's dead!) For more info see [this article](https://hackernoon.com/coffee-what-how-developers-quit-to-use-coffeescript-29aj36wn)
3. I could not found in the YouTube API documentation any guides, which used `gapi`

Instead of it I used request, which was made accordingly to the [API](https://developers.google.com/youtube/v3/docs/search/list).

Also I (will add) added `README.md`, which contains info about running application locally, deploying and running it in Cloud Shell. And there's also a short explanation why it's so complicated. @sashapff, @BudiGabe please have a look to it, because after merging this branch we'll need to use these instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,81 @@
 # Alpollo
 
 Capstone project for STEP 2020
+
+## Add sensitive data locally
+
+Add project ID to `src/main/webapp/WEB-INF/config.json`
+
+Add API-KEY and other sensitive data you need for Node JS to `src/main/webapp/.env`
+
+*** DO NOT PUSH THESE FILES TO GIT ***
+
+In case you did this fatal mistake: change API-key/Client ID in GCP and inform others about it and don't forget to do deploy with new keys.
+
+## Run locally
+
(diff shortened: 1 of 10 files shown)


jeinygroove:
I have a strange bug in deploy. Locally and in Cloud Shell everything works absolutely fine. Also everything was fine when I deployed first 2-3 times, but after it I started to get strange errors in [log](https://pantheon.corp.google.com/appengine?project=google.com:alpollo-step-2020)

At first it was an error when I did request to `/api/youtube` (`While handling this request, the process that handled this request was found to be using too much memory and was terminated`)

Then I even started to get the next things: `npm run server exited with code SIGTERM` and `npm run client exited with code SIGTERM`

PR #38, opened 2020-08-31
Added YouTube servlet

jeinygroove:
Since there's an unknown bug with #36, I wrote a YouTube servlet just in case.

### DIFF
--- a/src/main/java/com/google/alpollo/YouTubeService.java
+++ b/src/main/java/com/google/alpollo/YouTubeService.java
@@ -12,7 +12,7 @@ import java.security.GeneralSecurityException;
 /**
  * Class for creating YouTube service.
  */
-public class YouTubeService {
+public final class YouTubeService {
   private static final JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
 
   private YouTubeService() {}
--- a/src/main/java/com/google/alpollo/YouTubeService.java
+++ b/src/main/java/com/google/alpollo/YouTubeService.java
@@ -6,15 +6,11 @@ import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.services.youtube.YouTube;
 import com.google.api.services.youtube.YouTubeRequestInitializer;
-
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 
 public class YouTubeService {
   private static final JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
-
-  public YouTubeService() {
-  }
 
   /**
    * Build and return an authorized API client service.
--- a/src/main/java/com/google/alpollo/YouTubeService.java
+++ b/src/main/java/com/google/alpollo/YouTubeService.java
@@ -3,7 +3,7 @@ package com.google.alpollo;
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.JsonFactory;
-import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.services.youtube.YouTube;
 import com.google.api.services.youtube.YouTubeRequestInitializer;
 
@@ -11,7 +11,7 @@ import java.io.IOException;
 import java.security.GeneralSecurityException;
 
 public class YouTubeService {
-  private static final JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
+  private static final JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
 
   public YouTubeService() {
   }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/alpollo/YouTubeService.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public YouTubeService() {}
+
+    /**
+     * Build and return an authorized API client service.
+     *
+     * @return an authorized API client service
+     * @throws GeneralSecurityException, IOException
+     */
+    public static YouTube getService(String apiKey) throws GeneralSecurityException, IOException {
```
RazvanBudacaTechquarter: Spacing seems a bit big here. Did you use 4 spaces instead of 2?

src/main/java/com/google/alpollo/YouTubeServlet.java
```diff
@@ -0,0 +1,48 @@
...
+
+@WebServlet("/youtube")
+public class YouTubeServlet extends HttpServlet {
+  /**
+   * Accesses YouTube and retrieves a collection of videos relating to the given query/words found
+   * in the request. Returns only the snippet for each video and returns only videos, no playlists
+   * or channels allowed.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
RazvanBudacaTechquarter, line 26: Separating the code inside `doGet()` into smaller code-blocks with whitespaces might make the code easier to read.

RazvanBudacaTechquarter (approved):
image

src/main/java/com/google/alpollo/YouTubeService.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.api.services.youtube.YouTube;
+import com.google.api.services.youtube.YouTubeRequestInitializer;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+public class YouTubeService {
+  private static final JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
+
+  public YouTubeService() {
```
sashapff: You don't need it because java automatically generate this constructor.

src/main/java/com/google/alpollo/YouTubeService.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.api.services.youtube.YouTube;
+import com.google.api.services.youtube.YouTubeRequestInitializer;
+
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+public class YouTubeService {
```
sashapff: Add documentation to this and other classes.

src/main/java/com/google/alpollo/YouTubeService.java
```diff
@@ -0,0 +1,33 @@
...
+  private static final JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
```
jgaillard85: Is there a reason to use Jackson here and not Gson like in other parts of the project?

src/main/java/com/google/alpollo/YouTubeServlet.java
```diff
@@ -0,0 +1,51 @@
...
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Empty query.");
+      return;
+    }
+
+    YouTube youtubeService = null;
+    try {
+      youtubeService = YouTubeService.getService(ConfigHelper.getSensitiveData(this.getServletContext(), ConfigHelper.SENSITIVE_DATA.API_KEY));
+    } catch (Exception e) {
+      response.getWriter().println(gson.toJson(e));
+      e.printStackTrace();
```
jgaillard85: I am not sure that is the best way to log the exception here.

jeinygroove: You mean that it's better to use another methods? For example, something from `java.util.logging`.

jgaillard85: I just meant that printing the stack trace is not a recommended way of logging exceptions.
As I mentioned in another PR, at some point it would be interesting to figure out proper server-side logging, but that is not a priority.

src/main/java/com/google/alpollo/YouTubeService.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.api.services.youtube.YouTube;
+import com.google.api.services.youtube.YouTubeRequestInitializer;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+
+/**
+ * Class for creating YouTube service.
+ */
+public final class YouTubeService {
+  private static final JsonFactory jsonFactory = GsonFactory.getDefaultInstance();
```
jgaillard85: Use upper case for the names of constants

jgaillard85 (approved):

PR #39, opened 2020-08-31
Added Youtube component for YouTube Servlet

jeinygroove:


src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -0,0 +1,110 @@
...
+      );
+  }
+
+  render() {
+    const classes = this.props.classes;
+
+    if (this.state.error) {
+      return (
+        <div>
+          <p>{this.state.error.message}</p>
```
RazvanBudacaTechquarter: Again, the user might not know what this error means.

src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -0,0 +1,110 @@
...
+              />
+            </ListItem>
+          ))}
+        </List>
+      </div>
+    );
+  }
+
+  _onReady(event) {
+    // access to player in all event handlers via event.target
```
RazvanBudacaTechquarter: nit: Capitalize the first letter and full stop at the end.

src/main/webapp/src/components/youTubeRecommendations.js
jeinygroove: Usually it's "Internal Server error 500", I think there's no need to explain the user why it happened, because they shouldn't care about it)

RazvanBudacaTechquarter: Well, if the user shouldn't care about it, then it shouldn't be in the UI in the first place. I think it's a better idea to give the user easier to process information. Maybe change it to `Oops, something went wrong` as in the other request?

jeinygroove: Yep

src/main/webapp/src/components/songInfo.js
```diff
@@ -1,15 +1,14 @@
...
 import Typography from "@material-ui/core/Typography";
-import List from "@material-ui/core/List";
-import ListItem from "@material-ui/core/ListItem";
 import SentimentAnalysisInfo from "./sentimentAnalysisInfo";
 import EntityAnalysisInfo from "./entityAnalysisInfo";
+import YouTubeRecommendations from "./youTubeRecommendations";
 import Lyrics from "./lyrics";
 import { Redirect } from "react-router";
 
+
```
sashapff: nit: empty line

RazvanBudacaTechquarter (approved):

src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -0,0 +1,111 @@
...
+import YouTube from "react-youtube";
+import List from "@material-ui/core/List";
+import ListItem from "@material-ui/core/ListItem";
+import CircularProgress from '@material-ui/core/CircularProgress';
+import axios from "axios";
+
+const styles = (theme) => ({
+  youTubeVideo: {
+    playerVars: {
+      autoplay: 1,
```
jgaillard85: Maybe add a comment here to explain if that means autoplaying or not

jgaillard85 (approved):

PR #40, opened 2020-09-01
Create separate SongInfo class

RazvanBudacaTechquarter:
We created this class to separate the analysis results from the "superficial" data (artist name, song name), mostly because Sasha doesn't need to work with the analysis info.
This class will be used to store aaaaaall the info about songs in a single object and send it for storage.


src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -16,4 +17,14 @@ public static void saveSongRequest(Song song) {
   public static List<Song> topSongs() {
     return OfyService.ofy().load().type(Song.class).limit(TOP_SIZE).list();
   }
+
+  /** Save whole song info to database. */
+  public static void saveSongInfo(SongInfo songInfo) {
```
jeinygroove, line 22: I think you should also return the id of the song entity after saving it, since we'll need it for retrieving the song info from database.

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.alpollo.Song;
+import com.google.alpollo.model.SongEntity;
+import com.googlecode.objectify.annotation.Index;
+import com.googlecode.objectify.annotation.Entity;
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ * 
+ * We extend from Song because we do not always need the entire song data. Sasha mostly
```
jeinygroove: You don't need this comment in documentation

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,38 @@
...
+public class SongInfo extends Song {
+  /** Score tells us the overall positivty of the song, ranging from -1.0 to 1.0. */
+  @Index private float score;
+  /** Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf. */
+  @Index private float magnitude;
+  /** TopSalientEntities is a list containing the 10 most important words, given the song context. */
+  @Index private List<SongEntity> topSalientEntities;
+  @Index private String lyrics;
+  /** After analyzing the entities, we'll have a list of recommended YouTube links. */
+  @Index private List<String> youTubeLinks;
```
jeinygroove: It's not links, it's video ids

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,38 @@
...
+@Entity
+public class SongInfo extends Song {
+  /** Score tells us the overall positivty of the song, ranging from -1.0 to 1.0. */
+  @Index private float score;
+  /** Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf. */
+  @Index private float magnitude;
+  /** TopSalientEntities is a list containing the 10 most important words, given the song context. */
+  @Index private List<SongEntity> topSalientEntities;
+  @Index private String lyrics;
+  /** After analyzing the entities, we'll have a list of recommended YouTube links. */
```
jeinygroove: There's no need to explain in documentation how we made this recommendations, it's just smth like 
`List of recommended YouTube video IDs, based on song lyrics/entity analysis.`

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -16,4 +17,14 @@ public static void saveSongRequest(Song song) {
...
+    OfyService.ofy().save().entity(songInfo).now();
+  }
+
+  /** Returns the song info by song id. */
+  public static SongInfo getSongInfo(int id) {
```
jeinygroove: According to the documentation and Song class, id should have `String` type instead of `int`

RazvanBudacaTechquarter: I'll let Sasha handle this one.

jeinygroove (approved):

src/main/java/com/google/alpollo/model/SongInfo.java
jgaillard85: Do not mention names in the javadoc, and think whether what you write can be understood by someone with little to no context.

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * Extending the Song class, it provides all the info related to a specific song.
+ * 
+ * We extend from Song because we do not always need the entire song data. Sasha mostly
+ * needs to work with the "superficial" data (artist name, song name etc.) while Budi has to 
+ * work with the analysis results.
+ */
+@Entity
+public class SongInfo extends Song {
```
jgaillard85: Similarly to what I wrote in PR #26, this class inheritance does not make sense. With Object Oriented Programming, you need to think of a subclass as being a particular kind of its parent class. But SongInfo is not a Song. If you want to add data to a Song, then add fields inside the Song class.
If you want to keep all those together because they form so sort of logical block, then keep SongInfo (maybe with a more illuminating name), but let it just contain the fields, and not be a subclass of Song. Then Song can have a SongInfo field.
And if you want to still be able to have Song objects that don't have SongInfo, then you can have a constructor that sets this field to null, or set the value of this field through a setter instead of inside the constructor

jeinygroove:
Seems like I accidentally approved this pr, because I saw that comments were resolved and haven't noticed that there wasn't a commit. Idk why my comments were `marked as resolved by @BudiGabe` or it was made automatically by github. I made some of them unresolved and the first one about comment with names was also mentioned in Jerome's review.

RazvanBudacaTechquarter:
> Seems like I accidentally approved this pr, because I saw that comments were resolved and haven't noticed that there wasn't a commit. Idk why my comments were `marked as resolved by @BudiGabe` or it was made automatically by github. I made some of them unresolved and the first one about comment with names was also mentioned in Jerome's review.

Something happened and the commits got removed. Maybe it happened after we tried to get all the branches up to date. I remember fixing those issues.

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,36 @@
...
+  /** All the info in this class is attached to a specific parent song. */
+  private final Song parentSong;
+  /** Score tells us the overall positivty of the song, ranging from -1.0 to 1.0. */
+  @Index private float score;
+  /** Magnitude tells us how strong the main sentiment of the song is, ranging from 0.0 to +inf. */
+  @Index private float magnitude;
+  /** TopSalientEntities is a list containing the 10 most important words, given the song context. */
+  @Index private List<SongEntity> topSalientEntities;
+  @Index private String lyrics;
+  /** After analyzing the entities, we'll have a list of recommended YouTube links. */
```
jeinygroove: Change links to video ids in the documentation too

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.alpollo.model.SongEntity;
+import com.googlecode.objectify.annotation.Index;
+import com.googlecode.objectify.annotation.Entity;
+
+/**
+ * Class used for storing data related to the Natural Language API analysis.
+ * It provides all the info related to a specific song.
+ */
+@Entity
+public class SongInfo {
```
jgaillard85: I think a more explicit name would be better. Something like SentimentAnalysisInfo, something that shows what sort of data is in there. That might help convey better that this is not something to have for every song

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -0,0 +1,36 @@
...
+  /** All the info in this class is attached to a specific parent song. */
+  private final Song parentSong;
```
sashapff: Don't forget to rename `parentSong` to `song` as we discussed.

src/main/java/com/google/alpollo/SongDataBase.java
sashapff: I've done

src/main/java/com/google/alpollo/model/SongInfo.java
jeinygroove: This class doesn't contain `id` field, without it this code won't compile, because you register this class in objectify. @sashapff knows how to fix it

src/main/java/com/google/alpollo/OfyService.java
```diff
@@ -3,12 +3,14 @@
 import com.googlecode.objectify.Objectify;
 import com.googlecode.objectify.ObjectifyFactory;
 import com.googlecode.objectify.ObjectifyService;
+import com.google.alpollo.model.SongInfo;
 
 /** Helper class to start running Objectify. */
 public class OfyService {
   /** Registers Song class in Objectify. */
```
jgaillard85: Update documentation

sashapff: I fixed it in pr#26

sashapff (dismissed):

amaurymedeiros (approved):

PR #41, opened 2020-09-01
Synchronize all analysis components

jeinygroove:
Now youtube loads not static videos, but videos based on the most salient word, found by entity analysis

### DIFF
--- a/src/main/webapp/src/components/youTubeRecommendations.js
+++ b/src/main/webapp/src/components/youTubeRecommendations.js
@@ -57,33 +57,29 @@ class YouTubeRecommendations extends React.Component {
         error: new Error(entityState.errorMsg),
         isLoading: false,
       });
+    } else if (entityState.entityAnalysisInfo.length == 0) {
+      this.setState({
+        error: new Error("No YouTube videos were found, because no entities were found"),
+        isLoading: false,
+      });
     } else {
-      if (entityState.entityAnalysisInfo.length == 0) {
-        this.setState({
-          error: new Error(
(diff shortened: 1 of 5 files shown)


sashapff (approved):

src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -41,24 +42,66 @@ class YouTubeRecommendations extends React.Component {
...
+              this.setState({
+                error,
+                isLoading: false,
+              })
+            );
+        }
+      }
+    } else {
+      this.setState({ isLoading: true });
+    }
```
jgaillard85: For readability, I would switch the two conditions here, and it is usually better to have positive conditions if possible. So I would do
```
else if (entityState.isLoading) {
...
} else {
...
}
```

src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -41,24 +42,64 @@ class YouTubeRecommendations extends React.Component {
...
+            });
+          })
+          .catch((error) =>
+            this.setState({
+              error,
+              isLoading: false,
+            })
+          );
+      }
+    }
```
jgaillard85: Now you can even remove one level of indentation by having directly
```
else if (entityState.entityAnalysisInfo.length == 0) {
...
} else {
...
}
```

jgaillard85 (approved):

jgaillard85 (approved):

PR #43, opened 2020-09-02
Budi gathers data

RazvanBudacaTechquarter:
Making requests to this servlet will allow us to get or store SongInfo objects containing all the song data we want.


src/main/java/com/google/alpollo/SongInfoServlet.java
```diff
@@ -0,0 +1,45 @@
...
+  private static final String SONG_INFO = "songInfo";
+  private final Gson gson = new Gson();
+
+  /**
+   * Making a GET request to our servlet with the desired song ID as a parameter
+   * will return that song from the database.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    int songId = gson.fromJson(request.getParameter(SONG_ID), int.class);
```
jeinygroove: Same as in #40, id is `String`

src/main/java/com/google/alpollo/SongInfoServlet.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.alpollo.model.SongInfo;
+import com.google.gson.Gson;
+
+/** Servlet created to deal with songInfo related operations. */
```
jeinygroove: Please, enumerate these operations.

src/main/java/com/google/alpollo/SongInfoServlet.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Making a POST request to the server with the songInfo object as a parameter (object containing
+   * all the info related to our song) will send the song to the storage layer. 
+   */
+  @Override 
+  public void doPost(HttpServletRequest request, HttpServletResponse response) {
+    SongInfo songInfo = gson.fromJson(request.getParameter(SONG_INFO), SongInfo.class);
+    SongDataBase.saveSongInfo(songInfo);
```
jeinygroove: This method can throw an exception if something with objectify went wrong, so we should handle it and send the error in that case

src/main/java/com/google/alpollo/SongInfoServlet.java
RazvanBudacaTechquarter: Well each operation is described inside, above the doGet() method and the doPost() method.

jeinygroove: It's good that there's just 2 methods, but if they were more, it would be quite uncomfortable to scroll down through all of them((

jeinygroove: You can just add that this servlet `retrieves and saves songInfo object to the database`

src/main/java/com/google/alpollo/SongInfoServlet.java
RazvanBudacaTechquarter: I had no idea. Do you mean `saveSongInfo()`?

src/main/java/com/google/alpollo/SongInfoServlet.java
```diff
@@ -0,0 +1,45 @@
...
+  private final Gson gson = new Gson();
+
+  /**
+   * Making a GET request to our servlet with the desired song ID as a parameter
+   * will return that song from the database.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String songId = request.getParameter(SONG_ID);
+    SongInfo songInfo = SongDataBase.getSongInfo(songId);
```
jeinygroove: This string won't compile, because on this branch getSongInfo expects `int` instead of `String`. You can fix on `songInfo-class` branch and upload this changes to this branch

src/main/java/com/google/alpollo/SongInfoServlet.java
jeinygroove: I don't know objectify, I just saw that, for example, `ofy()` can throw `IllegalStateExpression`. I think you or @sashapff should read the documentation and see what can exceptions can occur.

I don't think that we are able to catch specific types of exception since they depend on objectify, so maybe it's enough just to put this code to `try/catch` block and send the exception message if it occurred.

jeinygroove (dismissed):

jeinygroove (commented):
Accidentally approved, idk how to undo it, looks like it's impossible..

src/main/java/com/google/alpollo/SongInfoServlet.java
sashapff: Why there you get the same id? I think in this case we need to use the same static method to get id from the song name and artist, but I didn't see it in the js code. It is more question to Olya. Why we have right id here?

src/main/java/com/google/alpollo/SongInfoServlet.java
sashapff: It's reasonable. I will fix the same issue in the TopSongs Servlet as well.

src/main/java/com/google/alpollo/model/SongInfo.java
```diff
@@ -23,16 +19,16 @@
   /** TopSalientEntities is a list containing the 10 most important words, given the song context. */
   @Index private List<SongEntity> topSalientEntities;
   @Index private String lyrics;
-  /** After analyzing the entities, we'll have a list of recommended YouTube links. */
-  @Index private List<String> youTubeLinks;
+  /** List of YouTube video IDs based on the most salient entity */
```
sashapff: Problem with the documentation here was in another pr but it's fine. Just drawn your attention.

src/main/java/com/google/alpollo/SongInfoServlet.java
RazvanBudacaTechquarter: Olya can do a get request here to get a song by it's id, wasn't that the idea? Maybe I didn't understand.

jeinygroove: We discussed it today, so I think that it's solved (if not, let's discuss it in chat)

sashapff (approved):

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  /**
+   * Making a POST request to the server with the analysisInfo object as a parameter (object containing
+   * all the info related to our song) will send the song to the storage layer. 
+   */
+  @Override 
+  public void doPost(HttpServletRequest request, HttpServletResponse response) {
+    AnalysisInfo analysisInfo = gson.fromJson(request.getParameter(ANALYSIS_INFO), AnalysisInfo.class);
```
jeinygroove: Remember that parameter is NOT in the body, it's in the query of the URL. You should take `request.getReader()` instead.

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
```diff
@@ -0,0 +1,50 @@
...
+   * parameter (object containing all the info related to our song) will send the
+   * song to the storage layer.
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    AnalysisInfo analysisInfo;
+    try {
+      analysisInfo = gson.fromJson(request.getReader(), AnalysisInfo.class);
+      SongDataBase.saveAnalysisInfo(analysisInfo);
+    } catch (JsonSyntaxException | JsonIOException | IOException e) {
```
jeinygroove, line 46: have a look at the Amaury comment in #49, please

jeinygroove (approved):

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
```diff
@@ -0,0 +1,50 @@
...
+public class AnalysisInfoServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+
+  /**
+   * Making a GET request to our servlet with the desired song ID as a parameter
+   * will return that song from the database.
+   */
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long songId = gson.fromJson(request.getReader(), Long.class);
```
jeinygroove: getReader reads the body of the request, GET requests can't have body, you should get the parameter here as you did before

RazvanBudacaTechquarter: Ok it makes sense now.

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
RazvanBudacaTechquarter: After removing the `throws IOException` from both methods, Cloud Shell refuses to compile and test, even though it's inside the catch line.

amaurymedeiros: The problem seems to be in `doGet`. Either `gson.fromJson` or `SongDataBase.getAnanlysisInfo` might throw `IOException` and are not inside the `try/catch` block. Can you try moving everything inside the try/catch and see if it compiles?

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -2,20 +2,23 @@
 
 import java.util.List;
 import com.google.alpollo.model.AnalysisInfo;
```
amaurymedeiros, line 4: Changes to this file were already merged in PR #51.
Can you please rebase this change to make it easier to review?

amaurymedeiros: nevermind. I realized the other PRs were merged into the branch of this PR. I'll review it now.

src/main/webapp/src/components/songInfo.js
```diff
@@ -94,18 +100,53 @@ class SongInfo extends React.Component {
...
+  }
+
+  /** Checks if the state of the component was loaded successfully. */
+  isReady(componentState) {
+      return (componentState && !componentState.isLoading && !componentState.errorMsg);
+  }
+
+  /** If all analysis components are ready, send SongInfo object to the server. */
+  sendSongInfo(state) {
+    if (!state.wasSent && this.isReady(state.youTubeState) && this.isReady(state.entityState) && this.isReady(state.sentimentState)) {
```
amaurymedeiros: It would probably be better to extract `this.isReady(state.youTubeState) && this.isReady(state.entityState) && this.isReady(state.sentimentState)` to a separate function, e.g. `isReady(state)`, so the if condition is better to read.

src/main/webapp/src/components/sentimentAnalysisInfo.js
```diff
@@ -50,27 +50,35 @@ class SentimentAnalysisInfo extends React.Component {
     }
   }
 
+  /** 
+   * Get sentiment analysis info if it was sent before
+   * or load it instead.
+   */
   componentDidMount() {
```
amaurymedeiros: Since `componentDidMount` is a react method called at some point in the lifecycle, I'd rather make it call a separate method, where you can place your documentation. That will escalate better when/if you need to do more actions in the `componentDidMount` function.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -35,27 +35,35 @@ class EntityAnalysisInfo extends React.Component {
     }
   }
 
+  /** 
+   * Get entity analysis info if it was sent before
+   * or load it instead.
+   */
   componentDidMount() {
```
amaurymedeiros: Since `componentDidMount` is a react method called at some point in the lifecycle, I'd rather make it call a separate method, where you can place your documentation. That will escalate better when/if you need to do more actions in the `componentDidMount` function.

amaurymedeiros (commented):
LGTM apart from minor comments

src/main/webapp/src/components/youTubeRecommendations.js
```diff
@@ -81,16 +81,28 @@ class YouTubeRecommendations extends React.Component {
...
+   * Get youtube video ids if they were sent before
+   * or load it instead.
+   */
   componentDidMount() {
-    this.getVideos(this.props.entityState);
+    if (this.props.sentInfo.wasSent) {
+      this.setState({ videoIds: this.props.sentInfo.info });
+    } else {
+      this.getVideos(this.props.entityState);
+    }
```
amaurymedeiros: Since `componentDidMount` is a method from the react lifeCycle, let's make it simply call another method instead of having a non-trivial logic. That might not make too much of a difference now, but it will if you want to do it expand the logic in the future. Then you can move the documentation to this method. The same was done, for instance, in `sentimentAnalysisInfo.js`

jgaillard85 (approved):

PR #44, opened 2020-09-02
Sentiment interpretation

RazvanBudacaTechquarter:
Based on the score and magnitude of the sentiment, now we're sending the interpretation of these value to the client through the SongSentiment object. This way, the user will understand way easier what those 2 values actually mean.

### DIFF
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -107,6 +107,15 @@ public final class AnalysisHelper {
     if(sentiment.getMagnitude() < 1) {
       return "Neutral";
     }
-    
+
+    if(sentiment.getScore() < -0.15f) {
+      return "Negative";
+    }
+
+    if(sentiment.getScore() > 0.15f) {
+      return "Positive";
+    }
+
+    return "Mixed";
   }
 }
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -110,16 +110,16 @@ public final class AnalysisHelper {
     }
   }
 
-  public static String interpretSentiment(SongSentiment sentiment) {
-    if(sentiment.getMagnitude() < NEUTRAL_THRESHOLD) {
+  public static String interpretSentiment(double score, double magnitude) {
+    if(magnitude < NEUTRAL_THRESHOLD) {
       return NEUTRAL;
     }
 
-    if(sentiment.getScore() < NEGATIVE_THRESHOLD) {
+    if(score < NEGATIVE_THRESHOLD) {
       return NEGATIVE;
     }
 
-    if(sentiment.getScore() > POSITIVE_THRESHOLD) {
+    if(score > POSITIVE_THRESHOLD) {
       return POSITIVE;
     }
 
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -109,6 +109,10 @@ public final class AnalysisHelper {
     }
   }
 
+  /**
+   * Given the score and magnitude of a sentiment, this will return a general interpretation,
+   * telling the user what the values mean.
+   */
   public static String getInterpretation(double score, double magnitude) {
     if(magnitude < NEUTRAL_THRESHOLD) {
       return NEUTRAL;
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -4,7 +4,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import com.google.alpollo.model.SongEntity;
-import com.google.alpollo.model.SongSentiment;
 import com.google.api.gax.rpc.FixedHeaderProvider;
 import com.google.cloud.language.v1.AnalyzeEntitiesRequest;
 import com.google.cloud.language.v1.AnalyzeEntitiesResponse;
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -28,7 +28,7 @@ public final class AnalysisHelper {
   private static final String NEGATIVE = "This song is predominantly negative.";
   private static final String POSITIVE = "This song is predominantly positive";
   private static final String MIXED = "This song has a lot of mixed feelings";
-  private static final double NEUTRAL_THRESHOLD = 1;
+  private static final double NEUTRAL_THRESHOLD = 5;
   private static final double NEGATIVE_THRESHOLD = -0.15;
   private static final double POSITIVE_THRESHOLD = 0.15;
 
--- a/src/main/java/com/google/alpollo/AnalysisHelper.java
+++ b/src/main/java/com/google/alpollo/AnalysisHelper.java
@@ -114,15 +114,15 @@ public final class AnalysisHelper {
    * telling the user what the values mean.
    */
   public static String getInterpretation(double score, double magnitude) {
-    if(magnitude < NEUTRAL_THRESHOLD) {
+    if (magnitude < NEUTRAL_THRESHOLD) {
       return NEUTRAL;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -24,6 +24,13 @@
   private static final int MAX_ENTITIES = 10;
   /** The index where we start our top list from. */
   private static final int FIRST_ENTITY = 0;
+  private static final String NEUTRAL = "There isn't much going in this song.";
```
jeinygroove: Maybe we can explain neutral in a little bit more formal way.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -24,6 +24,13 @@
...
+  private static final String NEGATIVE = "This song is predominantly negative.";
+  private static final String POSITIVE = "This song is predominantly positive";
+  private static final String MIXED = "This song has a lot of mixed feelings";
+  private static final double NEUTRAL_THRESHOLD = 1;
```
jeinygroove: Why have you chosen this threshold? For example, I saw the song which had ~36 magnitude, so I think that 1  is a very small number.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -24,6 +24,13 @@
...
+  private static final double NEGATIVE_THRESHOLD = -0.15;
```
jeinygroove: Can you add comments to this constants or rename them somehow, because it wasn't obvious that `NEUTRAL_THRESHOLD` is used for comparing with `magnitude`, but `NEGATIVE_THRESHOLD` is used with `score`.

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -101,4 +108,24 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
         return tempList.subList(FIRST_ENTITY, MAX_ENTITIES);
     }
   }
+
+  /**
+   * Given the score and magnitude of a sentiment, this will return a general interpretation,
+   * telling the user what the values mean.
+   */
+  public static String getInterpretation(double score, double magnitude) {
+    if(magnitude < NEUTRAL_THRESHOLD) {
```
jeinygroove: space after `if`

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -101,4 +108,24 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
...
+      return NEUTRAL;
+    }
+
+    if(score < NEGATIVE_THRESHOLD) {
```
jeinygroove: space after `if`

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -101,4 +108,24 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
...
+      return NEGATIVE;
+    }
+
+    if(score > POSITIVE_THRESHOLD) {
```
jeinygroove: space after `if`

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -55,6 +55,17 @@
...
+  private static final double NON_NEUTRAL_MAGNITUDE = 5;
+  private static final double NEUTRAL_SCORE = 0.1;
+  private static final double NEGATIVE_SCORE = -2;
+  private static final double POSITIVE_SCORE = 2;
+
+  private static final String NEUTRAL_INTERPRETATION = "There isn't much going in this song.";
+  private static final String NEGATIVE_INTERPRETATION = "This song is predominantly negative.";
+  private static final String POSITIVE_INTERPRETATION = "This song is predominantly positive";
+  private static final String MIXED_INTERPRETATION = "This song has a lot of mixed feelings";
+
```
jeinygroove: nit: extra empty line

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Neutral songs always have almost no magnitude. Maybe we could go a bit higher, maybe up to 2 or 3. Beyond that, we could consider there is something going on there.

jeinygroove: I think 5 is a little bit to much, maybe 2 will be fine. If not, we can change it later

jeinygroove (approved):

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -55,6 +55,16 @@
...
+  private static final double NEUTRAL_MAGNITUDE = 0;
+  private static final double NON_NEUTRAL_MAGNITUDE = 10;
+  private static final double NEUTRAL_SCORE = 0.1;
+  private static final double NEGATIVE_SCORE = -2;
+  private static final double POSITIVE_SCORE = 2;
+
+  private static final String NEUTRAL_INTERPRETATION = "There isn't much going in this song.";
+  private static final String NEGATIVE_INTERPRETATION = "This song is predominantly negative.";
+  private static final String POSITIVE_INTERPRETATION = "This song is predominantly positive";
+  private static final String MIXED_INTERPRETATION = "This song has a lot of mixed feelings";
```
jgaillard85: Reuse the strings from AnalysisHelper directly instead of redefining them, it is fine to have them public as they are constants.

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -160,4 +170,36 @@ public void checkEntitiesWithMetadata() throws IOException {
 
       Assert.assertThat(actual, CoreMatchers.is(expected));
   }
+
+  @Test
+  public void sentimentIsNeutral() {
+    String actual = AnalysisHelper.getInterpretation(NEUTRAL_SCORE, NEUTRAL_MAGNITUDE);
+    String expected = NEUTRAL_INTERPRETATION;
+
+    Assert.assertEquals(expected, actual);
```
jgaillard85: In here you could also test a non neutral score with a neutral magnitude

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -10,10 +10,13 @@
   private final float score;
   /** Each sentiment has a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf. */
   private final float magnitude;
+  /** Based on the score and magnitude of a sentiment, we can make up an interpretation of these values. */
+  private final String interpretation;
 
-  public SongSentiment(float score, float magnitude) {
+  public SongSentiment(float score, float magnitude, String interpretation) {
```
jgaillard85: Since the interpretation is not independent of the score and magnitude, I think it would make more sense to compute it inside the constructor instead of having it as a separate argument.

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: It should be impossible to have a non neutral score and neutral magnitude. Magnitude actual shows how much sentiment we have in the text somehow. Or at least the [documentation](https://cloud.google.com/natural-language/docs/basics#interpreting_sentiment_analysis_values) makes it seem so.

src/main/java/com/google/alpollo/model/SongSentiment.java
```diff
@@ -1,19 +1,24 @@
...
+ * useless properties of the Sentiment object. Class is mainly used for
+ * converting its instances to JSON strings.
  */
 public class SongSentiment {
   /** Each sentiment has a score, showing the overall positivity of the text, ranging from -1.0 to 1.0. */
   private final float score;
   /** Each sentiment has a magnitude, representing how strong the sentiment is, ranging from 0.0 to +inf. */
   private final float magnitude;
+  /** Based on the score and magnitude of a sentiment, we can make up an interpretation of these values. */
+  private final String interpretation;
```
jgaillard85, line 16: I am assuming you'll need a getter for this at some point. Better to add it in this PR as well if that's the case

jgaillard85 (approved):

PR #45, opened 2020-09-03
Send lyrics in body for POST requests

### LINKED ISSUE #35: Try to move parameters to body.
We need to find the way to send parameters in `body` with axios for POST requests. Now it appends all parameters to query, and if lyrics is too long we can get 414 error

jeinygroove:
Solves this issue #35.

### DIFF
--- a/src/main/java/com/google/alpollo/EntityServlet.java
+++ b/src/main/java/com/google/alpollo/EntityServlet.java
@@ -6,6 +6,7 @@ import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import com.google.alpollo.model.Lyrics;
 import com.google.alpollo.model.SongEntity;
 import com.google.cloud.language.v1.Entity;
 import com.google.gson.Gson;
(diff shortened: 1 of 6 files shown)


RazvanBudacaTechquarter (approved):

src/main/java/com/google/alpollo/EntityServlet.java
```diff
@@ -17,13 +18,11 @@
...
 public class EntityServlet extends HttpServlet {
   private final Gson gson = new Gson();
-  /** Parameter sent through the request which holds the song's lyrics. */
-  private static final String LYRICS = "lyrics";
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
     try {
-      String lyrics = request.getParameter(LYRICS);
+      final String lyrics = gson.fromJson(request.getReader(), Lyrics.class).getLyrics();
```
jgaillard85, line 25: Is it not possible to keep lyrics as String directly?
Something like 
`final String lyrics = gson.fromJson(request.getReader(), String.class);`
That would remove the need for the Lyrics class which is just a wrapper around String.

jeinygroove: Yes, we could, but in this case we need to change not only parsing of the body, but the format of request that we send from JSON to smth else, since single line JSON usually not considered as valid (if it's not array from 1 element, of course). So I thought that it would be better just to change parsing 🤔

jgaillard85 (approved):

PR #47, opened 2020-09-03
Find only music videos for recommendations

jeinygroove:
Usually the most popular videos are music clips, but sometimes it can show clickbaits/DIYs/other strange and not related videos.

This PR adds `topicId` parameter, which will help to find only videos related to music.


sashapff (approved):

src/main/java/com/google/alpollo/YouTubeServlet.java
```diff
@@ -17,6 +17,8 @@
  */
 @WebServlet("/youtube")
 public class YouTubeServlet extends HttpServlet {
+  private static final String MUSIC_TOPIC_ID = "/m/04rlf";
```
jgaillard85, line 21: Add a comment to explain where you got this value frome

jgaillard85 (approved):

PR #48, opened 2020-09-03
Chart is now clickable

### LINKED ISSUE #42: Find out if it's possible to display links in legend of charts.

RazvanBudacaTechquarter:
Pie elements that have a wiki link attached to them will show up a bit offset from the center. The user can click both the legend element and the pie segment. They will be redirected to the corresponding wiki page.

### DIFF
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -104,6 +104,13 @@ class EntityAnalysisInfo extends React.Component {
       }
     ];
 
+    var offsetSlices = [];
+    for(var i = 0; i < top10WordsLinks.length; i++) {
+      if(top10WordsLinks[i] !== "") {
+        offsetSlices[i] = {offset:0.1};
+      }
+    }
+
     return (
       <div>
         <div style={{ display: 'flex', maxWidth: 900 }}>
@@ -116,6 +123,7 @@ class EntityAnalysisInfo extends React.Component {
             options={{
               title: 'Most Important Words',
               legend: {viewWindow: { min: 0, max: 15 }},
+              slices: offsetSlices,
             }}
             chartEvents={chartEvents}
             legendToggle
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -97,7 +97,7 @@ class EntityAnalysisInfo extends React.Component {
           const [selectedItem] = chartWrapper.getChart().getSelection();
           const { row, column } = selectedItem;
 
-          if(top10WordsLinks[row] !== "") {
+          if (top10WordsLinks[row] !== "") {
             window.location.href = top10WordsLinks[row];
           }
         }
@@ -105,9 +105,9 @@ class EntityAnalysisInfo extends React.Component {
     ];
 
     var offsetSlices = [];
-    for(var i = 0; i < top10WordsLinks.length; i++) {
-      if(top10WordsLinks[i] !== "") {
-        offsetSlices[i] = {offset:0.1};
+    for (var i = 0; i < top10WordsLinks.length; i++) {
+      if (top10WordsLinks[i] !== "") {
+        offsetSlices[i] = { offset:0.1 };
       }
     }
 
@@ -122,7 +122,7 @@ class EntityAnalysisInfo extends React.Component {
             data={top10WordsData}
             options={{
               title: 'Most Important Words',
-              legend: {viewWindow: { min: 0, max: 15 }},
+              legend: { viewWindow: { min: 0, max: 15 } },
               slices: offsetSlices,
             }}
             chartEvents={chartEvents}
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -88,7 +88,7 @@ class EntityAnalysisInfo extends React.Component {
           + '<b>Salience: </b>' + entity.salience + '<br>' 
           + ((entity.wikiLink !== "") ? 'Click for wiki link' : '')
           + '</div>';
-  }
+    }
 
     /** Gather the most important words and their data in simple arrays. */
     var top10WordsData = [];
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -81,12 +81,21 @@ class EntityAnalysisInfo extends React.Component {
       return <p>No entities were found.</p>;
     }
 
+    /** Generate a custom tooltip for each entity */
+    function getTooltip(entity) {
+      return '<div style="font-size:15px">' + entity.name + '<br>' 
+          + '<b>Type </b>' + entity.type + '<br>' 
+          + '<b>Salience </b>' + entity.salience + '<br>' 
+          + ((entity.wikiLink !== "") ? 'Click for wiki link' : '')
+          + '</div>';
+  }
+
     /** Gather the most important words and their data in simple arrays. */
     var top10WordsData = [];
     var top10WordsLinks = [];
-    top10WordsData.push(['Word and Type', 'Importance'])
+    top10WordsData.push(['Word and Type', 'Importance', { role: "tooltip", type: "string", 'p': {'html': true} }])
     this.state.entityAnalysisInfo.forEach((entity) => {
-      top10WordsData.push([entity.name + ' (' + entity.type + ')', entity.salience]);
+      top10WordsData.push([entity.name + ' (' + entity.type + ')', entity.salience, getTooltip(entity)]);
       top10WordsLinks.push(entity.wikiLink);
     })
 
@@ -124,6 +133,8 @@ class EntityAnalysisInfo extends React.Component {
               title: 'Most Important Words',
               legend: { viewWindow: { min: 0, max: 15 } },
               slices: offsetSlices,
+              focusTarget: 'category',
+              tooltip: { isHtml: true }
             }}
             chartEvents={chartEvents}
             legendToggle
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -92,12 +92,6 @@ class EntityAnalysisInfo extends React.Component {
       }
     })
 
-    const listLinks = top10WordsLinks.map((link, index) =>
-      <li key={index + 1}>
-        <a href={link}> {link} </a>
-      </li>
-    );
-     
     return (
       <div>
         <div style={{ display: 'flex', maxWidth: 900 }}>
@@ -112,12 +106,6 @@ class EntityAnalysisInfo extends React.Component {
             }}
             legendToggle
           />
-        </div>
-        <div>
-        <Typography variant="h5">Wiki links found</Typography>
-          <ul>
-            {listLinks}         
-          </ul>
         </div>
       </div>
     );
--- a/src/main/webapp/src/components/entityAnalysisInfo.js
+++ b/src/main/webapp/src/components/entityAnalysisInfo.js
@@ -90,18 +90,31 @@ class EntityAnalysisInfo extends React.Component {
           + '</div>';
     }
 
+    /** Array that holds the data that will be displayed by the chart. */
     var top10WordsData = [];
+    /** Array that will hold the wiki links for later use. */
     var top10WordsLinks = [];
     /** 
      * Create a custom tooltip to overwrite the default one, we want to show whether a wiki link
      * is available, besides the default data.
-     * This tooltip will be able to interpret HTML, so we can format the text.
+     * The p tells the tooltip to have the HTML property, so we can format the text.
+     * https://developers.google.com/chart/interactive/docs/customizing_tooltip_content#customizing-html-content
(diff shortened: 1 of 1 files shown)


jeinygroove:
can you deploy it, please?

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -87,17 +87,30 @@ class EntityAnalysisInfo extends React.Component {
     top10WordsData.push(['Word and Type', 'Importance'])
     this.state.entityAnalysisInfo.forEach((entity) => {
       top10WordsData.push([entity.name + ' (' + entity.type + ')', entity.salience]);
-      if (entity.wikiLink !== "") {
-        top10WordsLinks.push(entity.wikiLink);
-      }
+      top10WordsLinks.push(entity.wikiLink);
```
jeinygroove, line 122: why do you removed if statement here and moved it to another place?

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -87,17 +87,30 @@ class EntityAnalysisInfo extends React.Component {
...
+
+          if(top10WordsLinks[row] !== "") {
+            window.location.href = top10WordsLinks[row];
+          }
+        }
+      }
+    ];
+
+    var offsetSlices = [];
+    for(var i = 0; i < top10WordsLinks.length; i++) {
```
jeinygroove: space after `for`

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -87,17 +87,30 @@ class EntityAnalysisInfo extends React.Component {
...
+      if(top10WordsLinks[i] !== "") {
```
jeinygroove: space after `if`

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -87,17 +87,30 @@ class EntityAnalysisInfo extends React.Component {
...
+        offsetSlices[i] = {offset:0.1};
```
jeinygroove: spaces: `{ offset: 0.1 }`

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -109,16 +122,13 @@ class EntityAnalysisInfo extends React.Component {
             data={top10WordsData}
             options={{
               title: 'Most Important Words',
+              legend: {viewWindow: { min: 0, max: 15 }},
```
jeinygroove: spaces: `{ viewWindow: { min: 0, max: 15 } }`

sashapff (approved):
LGTM, but still wait for deployment

jeinygroove (commented):
The problem is that even if parts of chart with offset are visible and noticeable, they don't look like link (maybe for some users they don't look clickable). Links should look like [links](). Have you tried to make words in legend look like [that]() or if it's not possible maybe we can make this parts with links look like button (make them more volume?)

I found [this doc](https://developers.google.com/chart/interactive/docs/customizing_tooltip_content) for you, so you can customize tooltip of the slice and make it contain a link, for example. 

Or I also saw this chart, so maybe it'll be also good to add a link at the bottom of the tooltip if it exists and call it `Wiki Link`
Screen Shot 2020-09-03 at 22 21 51

src/main/webapp/src/components/entityAnalysisInfo.js
RazvanBudacaTechquarter: I'm saving all wikiLinks, even if they are empty, to make sure the row and the array index are corresponding.

jeinygroove: got it

RazvanBudacaTechquarter:
> The problem is that even if parts of chart with offset are visible and noticeable, they don't look like link (maybe for some users they don't look clickable). Links should look like links. Have you tried to make words in legend look like that or if it's not possible maybe we can make this parts with links look like button (make them more volume?)
> 
> I found [this doc](https://developers.google.com/chart/interactive/docs/customizing_tooltip_content) for you, so you can customize tooltip of the slice and make it contain a link, for example.
> 
> Or I also saw this chart, so maybe it'll be also good to add a link at the bottom of the tooltip if it exists and call it `Wiki Link`
> Screen Shot 2020-09-03 at 22 21 51

It seems like the tooltip will make all the entries require links attached to them. If the link is an empty string, that could be bad for us.

> The problem is that even if parts of chart with offset are visible and noticeable, they don't look like link (maybe for some users they don't look clickable). Links should look like links. Have you tried to make words in legend look like that or if it's not possible maybe we can make this parts with links look like button (make them more volume?)
> 
> I found [this doc](https://developers.google.com/chart/interactive/docs/customizing_tooltip_content) for you, so you can customize tooltip of the slice and make it contain a link, for example.
> 
> Or I also saw this chart, so maybe it'll be also good to add a link at the bottom of the tooltip if it exists and call it `Wiki Link`
> Screen Shot 2020-09-03 at 22 21 51

Words in the legend do not support HTML so we can't really make them links. To add volume, I can only make the whole chart 3D, but not separate slices. 
Tooltips are customizable, but they're completely replaced, that might be the best solution.
The chart you found has the small tooltip added with chart.setAction() which sets the same tooltip to all slices, so it's not that useful to us.

jeinygroove:
> > The problem is that even if parts of chart with offset are visible and noticeable, they don't look like link (maybe for some users they don't look clickable). Links should look like links. Have you tried to make words in legend look like that or if it's not possible maybe we can make this parts with links look like button (make them more volume?)
> > I found [this doc](https://developers.google.com/chart/interactive/docs/customizing_tooltip_content) for you, so you can customize tooltip of the slice and make it contain a link, for example.
> > Or I also saw this chart, so maybe it'll be also good to add a link at the bottom of the tooltip if it exists and call it `Wiki Link`
> > Screen Shot 2020-09-03 at 22 21 51
> 
> It seems like the tooltip will make all the entries require links attached to them. If the link is an empty string, that could be bad for us.
> 
> > The problem is that even if parts of chart with offset are visible and noticeable, they don't look like link (maybe for some users they don't look clickable). Links should look like links. Have you tried to make words in legend look like that or if it's not possible maybe we can make this parts with links look like button (make them more volume?)
> > I found [this doc](https://developers.google.com/chart/interactive/docs/customizing_tooltip_content) for you, so you can customize tooltip of the slice and make it contain a link, for example.
> > Or I also saw this chart, so maybe it'll be also good to add a link at the bottom of the tooltip if it exists and call it `Wiki Link`
> > Screen Shot 2020-09-03 at 22 21 51
> 
> Words in the legend do not support HTML so we can't really make them links. To add volume, I can only make the whole chart 3D, but not separate slices.
> Tooltips are customizable, but they're completely replaced, that might be the best solution.
> The chart you found has the small tooltip added with chart.setAction() which sets the same tooltip to all slices, so it's not that useful to us.

Got it. So I can suggest to add customized tooltips, and like in the example with London olympiad you can create a function, which will receive the name of the word, it's salience and wikiLink, and return html with all the info if the link isn't empty or just standard tooltip info if it the link is empty.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,12 +81,21 @@ class EntityAnalysisInfo extends React.Component {
       return <p>No entities were found.</p>;
     }
 
+    /** Generate a custom tooltip for each entity */
+    function getTooltip(entity) {
+      return '<div style="font-size:15px">' + entity.name + '<br>' 
+          + '<b>Type </b>' + entity.type + '<br>'
```
jeinygroove: nit: `Type:` instead of `Type`

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,12 +81,21 @@ class EntityAnalysisInfo extends React.Component {
       return <p>No entities were found.</p>;
     }
 
+    /** Generate a custom tooltip for each entity */
+    function getTooltip(entity) {
+      return '<div style="font-size:15px">' + entity.name + '<br>' 
+          + '<b>Type </b>' + entity.type + '<br>' 
+          + '<b>Salience </b>' + entity.salience + '<br>'
```
jeinygroove: nit: `Salience:` instead of `Salience`

jeinygroove (approved):

src/main/webapp/src/components/entityAnalysisInfo.js
RazvanBudacaTechquarter: Damn, you were fast on this one.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,23 +81,45 @@ class EntityAnalysisInfo extends React.Component {
...
+            window.location.href = top10WordsLinks[row];
+          }
+        }
+      }
+    ];
+
+    var offsetSlices = [];
+    for (var i = 0; i < top10WordsLinks.length; i++) {
+      if (top10WordsLinks[i] !== "") {
+        offsetSlices[i] = { offset:0.1 };
```
amaurymedeiros: nit: space after ":"

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,23 +81,45 @@ class EntityAnalysisInfo extends React.Component {
...
+    function getTooltip(entity) {
+      return '<div style="font-size:15px">' + entity.name + '<br>' 
+          + '<b>Type: </b>' + entity.type + '<br>' 
+          + '<b>Salience: </b>' + entity.salience + '<br>' 
+          + ((entity.wikiLink !== "") ? 'Click for wiki link' : '')
+          + '</div>';
+    }
+
     /** Gather the most important words and their data in simple arrays. */
     var top10WordsData = [];
```
amaurymedeiros, line 94: the `top10WordsData` does not look simple to me. We add things like `Word and Type`, but also a subarray with the entity name and also type salience. Also, in the line below, we add `{ role: "tooltip", type: "string", 'p': {'html': true} }`, which is even less clear to me. Can you please add some documentation here explaining what this array should represent? Also, it might be worth to extract some things to variables with a meaningful name, such as the object above (`'role: "tooltip", type...`)

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,23 +81,45 @@ class EntityAnalysisInfo extends React.Component {
...
-        <a href={link}> {link} </a>
-      </li>
-    );
-     
+    const chartEvents = [
+      {
+        eventName: "select",
+        callback({ chartWrapper }) {
+          const [selectedItem] = chartWrapper.getChart().getSelection();
+          const { row, column } = selectedItem;
```
amaurymedeiros: you're not using `column`, so `{ row }` should be enough, right?

src/main/webapp/src/components/entityAnalysisInfo.js
RazvanBudacaTechquarter: Hmm yea, it used to be simpler in the old days.

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,23 +81,55 @@ class EntityAnalysisInfo extends React.Component {
...
+
     var top10WordsData = [];
     var top10WordsLinks = [];
-    top10WordsData.push(['Word and Type', 'Importance'])
+    /** 
+     * Create a custom tooltip to overwrite the default one, we want to show whether a wiki link
+     * is available, besides the default data.
+     * This tooltip will be able to interpret HTML, so we can format the text.
+     */
+    const customTooltip = { role: "tooltip", type: "string", 'p': {'html': true} };
```
amaurymedeiros: what is `p` and why do we need the single quotes only for it?

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -81,23 +81,55 @@ class EntityAnalysisInfo extends React.Component {
...
+
+    /** 
+     * Gather entity data in a separate array to display later in the chart.
```
amaurymedeiros: It's still not straightforward to me how this array looks like. When someone is reading the code, they should understand it immediately with little efforts. After reading the code for a bit, here is my interpretation:

`top10WordsData`'s elements should be arrays of size 3, where positions are:

0. String containing entity name + type
1. Number representing the entity salience
2. String containing an HTML `div` element representing the entity tooltip

However, the first element of `top10WordsData` is also an array of 3 positions which before this changed seemed to me that contained the "title" of each position. That's already a little bit odd to me (I'd rethink the structure if you need title + values), but I would understand it if React Charts, or other lib, requires the data in such format. In any case, this should be clear in the code and/or the inline document supporting the explanation. But after this change, we added the `customTooltip` object in the first subarray, which is not a string, making it even more confusing. Can you please clarify it?

src/main/webapp/src/components/entityAnalysisInfo.js
RazvanBudacaTechquarter: I must say that's just how I found it in multiple sources. Apparently you can add single quotes to the entire object. The p element tells the tooltip to interpret the tooltip text as HTML instead of the default interpretation.

amaurymedeiros (approved):
Thanks for the detailed documentation!

PR #49, opened 2020-09-07
Send song info

jeinygroove:
This branch combines code from #43 and #26, that's why it has extra commits from #26. I changed Song*.java and some servlets classes just to make my frontend part workable. Now we will send song info object when all analysis finishes successfully and after it we also save it to localStorage, it's important, because when the users refresh the page we won't do the same api requests (for youtube api it's essential, because it has the restriction to the number of search requests).

Rating list displays top songs as well, but it will display all searched songs because at this moment on backend we save all SongInfo's


jeinygroove:
ATTENTION: Today YouTube API won't work, because the limit of requests per day was exceeded :( You can see how it works and do code review tomorrow

sashapff (approved):

src/main/webapp/src/components/entityAnalysisInfo.js
```diff
@@ -35,27 +35,35 @@ class EntityAnalysisInfo extends React.Component {
     }
   }
 
+  /** 
+   * Get entity analysis info if it was sended before
```
jgaillard85: It is "sent", not "sended".
Fix in all the places you have used "sended"

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
```diff
@@ -43,6 +43,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     try {
       analysisInfo = gson.fromJson(request.getReader(), AnalysisInfo.class);
       SongDataBase.saveAnalysisInfo(analysisInfo);
+      SongDataBase.saveSongRequest(analysisInfo.getSong());
     } catch (JsonSyntaxException | JsonIOException | IOException e) {
       response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, e.getMessage());
```
amaurymedeiros, line 48: You're catching the exceptions here to respond with an error, which is the correct thing to do. You can remove the `throws` declaration from the method signature. Same for `doGet`

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -2,6 +2,9 @@
 
 import java.util.List;
 import com.google.alpollo.model.AnalysisInfo;
+import com.google.alpollo.model.Song;
+import com.google.alpollo.model.SongCounter;
+import java.util.ArrayList;
```
amaurymedeiros: You don't seem to add anything that needs ArrayList. Do you need this import?

src/main/java/com/google/alpollo/model/AnalysisInfo.java
```diff
@@ -15,7 +15,7 @@
 public class AnalysisInfo {
   @Id private Long id;
   /** All the info in this class is attached to a specific parent song. */
-  private final Song song;
+  @Index private Song song;
```
amaurymedeiros: Why did you remove `final`? Is this a requirement for using `@Index`?

src/main/webapp/src/components/songInfo.js
```diff
@@ -94,18 +100,53 @@ class SongInfo extends React.Component {
...
 
+  /** Save state of the component to localStorage. */
+  saveState() {
+    const json = JSON.stringify(this.state);
+    localStorage.setItem("state", json);
+  }
+
+  /** Checks if the state of the component was loaded successfully. */
+  isReady(componentState) {
+      return (componentState && componentState.isLoading === false && componentState.errorMsg === null);
```
amaurymedeiros: wouldn't `!componentState.isLoading` instead of `componentState.isLoading === false` work?

src/main/java/com/google/alpollo/model/AnalysisInfo.java
jeinygroove: Yes, objectify don't save `final` fields

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
jeinygroove: Redirecting it to @BudiGabe for #43

src/main/webapp/src/components/songInfo.js
jeinygroove: hmmm, yes it will work, in this method I was just expecting that state object will definitely have `isLoading` and `errorMsg` fields, that's why I made this check more concrete, but now I see that it's not necessary because of te default state of components, so I'll change it

jgaillard85 (approved):

src/main/java/com/google/alpollo/AnalysisInfoServlet.java
RazvanBudacaTechquarter: Hmm VSCode told me I should throw the IOException too, maybe it's wrong this time.

PR #50, opened 2020-09-08
Remove duplicate entities

### LINKED ISSUE #46: Duplicated entities
Sometimes Entity Analysis API returns equal words (e.g. LP - Lost on you returns `everything` as 3 different entities) 
Screen Shot 2020-09-03 at 17 47 10

We can delete less salient duplicates, for example.

RazvanBudacaTechquarter:
All duplicate entities have been combined into a single entity with multiple types. After this change, the API decided to change the order of entities with the same salience, so I had to chance the tests a bit. It should be all good though.

Also, instead of type NUMBER, we see UNRECOGNIZED, I don't understand why. The Natural Language API demo clearly shows "one" as being PERSON and NUMBER, but to us it return PERSON and UNRECOGNIZED.


src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +129,36 @@ public static String getInterpretation(double score, double magnitude) {
...
+    }
+
+    return new ArrayList<SongEntity>(map.values());
+  }
+
+  /** Method to avoid adding duplicate types to the same entity. */
+  private static boolean entityDoesNotHaveType(SongEntity entity, String newType) {
+    String[] types = entity.getType().split(", ");
+
+    for (String type: types) {
```
sashapff: nit: space before :

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -64,4 +64,20 @@ public int hashCode() {
...
+  public String getType() {
+    return type;
+  }
+
+  /** 
+   * Method to add a new type to the existing entity, removing the need of duplicate
+   * entities which have the same name but different types.
+   */
+  public void addType(String newType) {
+    this.type += ", " + newType;
```
sashapff: Change type from String to List of Strings. In this case you don't need to create splitter and split then.

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -17,7 +17,7 @@
    * All types can be seen here: 
    * https://cloud.google.com/natural-language/docs/reference/rest/v1/Entity#type
    */ 
-  private final String type;
+  private String type;
```
jeinygroove: It's better to make Set of strings since you can have multiple types for one entity, instead of concatenating all types to one string and make it not flexible for usage

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +129,36 @@ public static String getInterpretation(double score, double magnitude) {
...
+      } else {
+        map.put(entity.getName(), entity);
+      }
+    }
+
+    return new ArrayList<SongEntity>(map.values());
+  }
+
+  /** Method to avoid adding duplicate types to the same entity. */
+  private static boolean entityDoesNotHaveType(SongEntity entity, String newType) {
```
jeinygroove: If you'll have set of types, you won't need this method

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -90,7 +93,8 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
     for (Entity entity : entityList) {
       // Use round() here to set the double to 2 decimals.
       SongEntity simplifiedEntity = new SongEntity(entity.getName(), 
-          Math.round(entity.getSalience() * 100.0) / 100.0, entity.getType().toString(),
+          Math.round(entity.getSalience() * 100.0) / 100.0, 
+          new HashSet<String>(Arrays.asList(entity.getType().toString())),
```
jeinygroove: What about changing `SongEntity` `type` field to the list of types as well, so we'll be even more flexible :D

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +132,22 @@ public static String getInterpretation(double score, double magnitude) {
 
     return MIXED;
   }
+
+  public static List<SongEntity> getListWithoutDuplicates(List<SongEntity> list) {
+    HashMap<String, SongEntity> map = new HashMap<>(); 
+
+    for (SongEntity entity: list) {
```
sashapff: nit: space before `:`

sashapff (approved):

sashapff (approved):

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: The `type` field for `SongEntity` is `HashSet<String>` for now. In that line of code I'm creating the default SongEntity with a single type. What do you mean changing to list of types?

jeinygroove: You convert your `HashSet<String>` to one `String` and save it as a `type` in `SongEntity` object. We send this object to SongInfo, frontend, anywhere. So it doesn't make sense and it's not flexible, nobody will be able to operate with this types without parsing string

jeinygroove: So I mean that it will be good to change type of `type` field in `SongEntity` class

RazvanBudacaTechquarter: Hmm take a look at the tests, type is a Set of multiple strings. Right there I'm creating a Set with only 1 string because all entities have only one type in the beginning.

jeinygroove: ok, now I see it, got it, thanks)

jeinygroove (approved):

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +136,22 @@ public static String getInterpretation(double score, double magnitude) {
 
     return MIXED;
   }
+
+  public static List<SongEntity> getListWithoutDuplicates(List<SongEntity> list) {
```
amaurymedeiros: nit: maybe `filterDuplicates` would be a better/simpler name?

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +136,22 @@ public static String getInterpretation(double score, double magnitude) {
...
+    HashMap<String, SongEntity> map = new HashMap<>();
```
amaurymedeiros: You don't actually need the map to be `String` -> `SongEntity`, right? Probably `String` -> `Set<String>` would be enough.

Then in the for loop below you could simply do something like below:

``` java
Set<String> types = map.getOrDefault(entity.getName(), new HashSet<>());
types.addAll(entity.getType());
map.put(entity.getName(), types);
```

or, alternatively:
``` java
map.putIfAbsent(entity.getName(), new HashSet<>());
map.get(entity.getName()).addAll(entity.getType());
```

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -81,16 +84,21 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
...
+   * with only one string.
    */
   public static List<SongEntity> getSimplifiedEntityList(List<Entity> entityList) {
     List<SongEntity> simplifiedEntityList = new ArrayList<>();
 
     for (Entity entity : entityList) {
       // Use round() here to set the double to 2 decimals.
       SongEntity simplifiedEntity = new SongEntity(entity.getName(), 
-          Math.round(entity.getSalience() * 100.0) / 100.0, entity.getType().toString(),
+          Math.round(entity.getSalience() * 100.0) / 100.0,
```
amaurymedeiros: Should we round only after dividing by 100 instead?

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -16,12 +17,14 @@
    * Type matches the type of the entities generated by the Natural Language API.
    * All types can be seen here: 
    * https://cloud.google.com/natural-language/docs/reference/rest/v1/Entity#type
+   * 
+   * Each entity will have multiple, non-duplicate types.
    */ 
-  private final String type;
+  private HashSet<String> type;
```
amaurymedeiros: We only care about the interface type here. `HashSet` is not the only implementation of `Set` (other examples are `LinkedHashSet` and `TreeSet`), and we should use `Set` instead. This helps to achieve polymorphism and in general might help you when writing tests. For example, your tests could have your own implementation of the `Set` interface, and you could have more control over it. This might not be obvious now, but when you implement more complex structures in the future you'll definitely find it useful to have a production vs test implementation of an interface.

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -64,4 +67,12 @@ public int hashCode() {
     hash = 31 * hash + (wikiLink == null ? 0 : Objects.hashCode(wikiLink));
     return hash;
   }
+
+  public String getName() {
+    return name;
+  }
+
+  public HashSet<String> getType() {
```
amaurymedeiros: Same thing here. Returning `Set` should be enough.

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -124,11 +125,11 @@ public void top10SalientEntitiesWithLessThan10Entities() throws IOException {
     String responseString = responseWriter.toString();
 
     List<SongEntity> actual = gson.fromJson(responseString, new TypeToken<List<SongEntity>>(){}.getType());
-    List<SongEntity> expected = Arrays.asList(new SongEntity("mountain", 0.84, "OTHER", EMPTY_STRING),
-        new SongEntity("mountain", 0.06, "LOCATION", EMPTY_STRING),
-        new SongEntity("story", 0.05, "WORK_OF_ART", EMPTY_STRING),
-        new SongEntity("sale", 0.03, "OTHER", EMPTY_STRING),
-        new SongEntity("tale", 0.01, "WORK_OF_ART", EMPTY_STRING));
+    List<SongEntity> expected = Arrays.asList(
+        new SongEntity("mountain", 0.84, new HashSet<String>(Arrays.asList("OTHER", "LOCATION")), EMPTY_STRING),
```
amaurymedeiros: Shouldn't we also combine the salience here? In this case I'd expect to see  `{ "mountain", ["OTHER", "LOCATION"], 0.90 }`

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -166,8 +168,9 @@ public void checkEntitiesWithMetadata() throws IOException {
...
 
     List<SongEntity> actual = gson.fromJson(responseString, new TypeToken<List<SongEntity>>(){}.getType());
-    List<SongEntity> expected = Arrays.asList(new SongEntity("Google", 0.88, "ORGANIZATION", WIKI_LINK_GOOGLE),
-        new SongEntity("results", 0.12, "OTHER", EMPTY_STRING));
+    List<SongEntity> expected = Arrays.asList(
+        new SongEntity("Google", 0.88, new HashSet<String>(Arrays.asList("ORGANIZATION")), WIKI_LINK_GOOGLE),
+        new SongEntity("results", 0.12, new HashSet<String>(Arrays.asList("OTHER")), EMPTY_STRING));
 
       Assert.assertThat(actual, CoreMatchers.is(expected));
   }
```
amaurymedeiros, line 180: Let's add another test where the API returns the same type for the same entity, as I saw an occurrence of that happening in production, i.e.:

```
{ "mountain", "OTHER", 0.04 }
{ "mountain", "OTHER", 0.03 }
```

I'd expect to combine this results

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Hmm I see your point, but in this way I'm no longer returning a list of modified SongEntities. Will the sets of types remain modified in the objects in this way?

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: I don't understand what you mean here. First we multiply by 100 so the method rounds to the 2nd decimal, then after rounding we divide by 100 so we're back to single digit numbers, with 2 decimals.

For example, if we have a salience like 0.25674..., first we go to 25.6, round it to 26.0, then we go back to 0.26.

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: That's actually a great idea, we should add the saliences too.

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: I see what you're doing now. If it's just about formatting, you can simply use `DecimalFormat`.

``` java
private static DecimalFormat salienceFormat = new DecimalFormat("0.00");
...

new SongEntity(..., salienceFormat.format(entity.getSalience()), ...)

```

By default, `DecimalFormat` uses `RoundingMode.HALF_EVEN`, which uses the nearest neighbour when rounding, but you can force it to round up/down by calling `salienceFormat.setRoundingMode(RoundingMode.UP)` or `salienceFormat.setRoundingMode(RoundingMode.DOWN)`

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: d'oh, I missed the fact you need to return a `SongEntity` `List`. Never mind then!

src/test/java/com/google/alpollo/AnalysisTest.java
jeinygroove: @BudiGabe don't forget to change frontend part in this case and display according to the new format of response. Or I can do it if you want.

RazvanBudacaTechquarter: The response should be the same. What do you think there is to change?

jeinygroove: I just checked how it displays list of types, looks ok

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +139,25 @@ public static String getInterpretation(double score, double magnitude) {
...
+    for (SongEntity entity : list) {
+      if (map.containsKey(entity.getName())) {
+        SongEntity existingEntity = map.get(entity.getName());
+        Set<String> newTypes = entity.getType();
+        Set<String> existingTypes = existingEntity.getType();
+        double newSalience = entity.getSalience() + existingEntity.getSalience();
+        
+        existingTypes.addAll(newTypes);
+        // Apparently 0.84 + 0.06 = 0.899999999, so we have to format the salience again.
+        existingEntity.setSalience(Double.parseDouble(salienceFormat.format(newSalience)));
```
jgaillard85: I think the setSalience method should be the one responsible for the formatting, as future users might not know they are supposed to pass an already formatted number as argument to the method.
It could even be done in the getSalience method, as this only matters for display

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -81,16 +87,21 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
...
    */
   public static List<SongEntity> getSimplifiedEntityList(List<Entity> entityList) {
     List<SongEntity> simplifiedEntityList = new ArrayList<>();
 
     for (Entity entity : entityList) {
       // Use round() here to set the double to 2 decimals.
       SongEntity simplifiedEntity = new SongEntity(entity.getName(), 
-          Math.round(entity.getSalience() * 100.0) / 100.0, entity.getType().toString(),
+          Double.parseDouble(salienceFormat.format(entity.getSalience())), 
+          new HashSet<String>(Arrays.asList(entity.getType().toString())),
```
jgaillard85: To make this a bit cleaner, you could use the method Sets.newHashSet(...) from Guave

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: I for one think we should not format the salience on the backend at all. Do we have a good reason for it? In my opinion, the frontend should get the salience as it is and format it to 2 decimal places.

jgaillard85: Yes, I agree with Amaury here actually, it would be much better to have the front end be the one responsible for formatting the data it displays

RazvanBudacaTechquarter: If I don't format the salience in the backend, it's going to be impossible to test. All the salience values will have like 10 decimals and we can find only the first 2. Can I somehow set a tolerance?

jeinygroove: You can test them for not equality, but almost equality, just set error which will be ok, like 0.001, for example

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -53,7 +59,7 @@ public boolean equals(Object obj) {
   } 
 
     SongEntity entity = (SongEntity) obj;
-    return salience == entity.salience
+    return DoubleMath.fuzzyEquals(salience, entity.salience, TOLERANCE)
```
jgaillard85: This breaks the contract between equals and hashCode, as with this two entities can be equal but have different hash codes

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -128,4 +137,24 @@ public static String getInterpretation(double score, double magnitude) {
...
+        double newSalience = entity.getSalience() + existingEntity.getSalience();
+        
+        existingTypes.addAll(newTypes);
+        existingEntity.setSalience(newSalience);
+      } else {
+        map.put(entity.getName(), entity);
+      }
+    }
+
+    return new ArrayList<SongEntity>(map.values());
```
jgaillard85: I think there are too many list creations in this whole story, which is not great for performance. Here you create a list from the map.values() collection. But as far as I can tell, this list is only used as an input to getTopSalientEntities, which makes a copy of its input list as the first thing it does.
At least one of those copies seems unnecessary

RazvanBudacaTechquarter: Yep you're right. Sincerely, I have no idea why I'm making a copy in getTopSalientEntities().

src/main/java/com/google/alpollo/AnalysisHelper.java
```diff
@@ -81,31 +84,35 @@ public static Sentiment getSentiment(String projectID, String lyrics) throws Ill
...
       simplifiedEntityList.add(simplifiedEntity);
     }
 
     return simplifiedEntityList;
   }
 
-  public static List<SongEntity> getTopSalientEntities(List<SongEntity> simplifiedEntityList) {
-    List<SongEntity> tempList = new ArrayList<>(simplifiedEntityList);
-    tempList.sort(SongEntity.ORDER_BY_SALIENCE_DESCENDING);
+  public static List<SongEntity> getTopSalientEntities(List<SongEntity> filteredList) {
```
amaurymedeiros: Please add a documentation to this method that you modify the list passed as parameter. That might not be obvious to the caller.

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: I think you were copying it to avoid modifying the list passed as a parameter. I personally wouldn't expect a function called `getSomethingBlaBla(myList)` to modify `myList`.

If we're only using these two functions together, the best thing to do is probably merge them (remember to rename the function accordingly). So instead of returning `new ArrayList<SongEntity>(map.values())` here you'd already sort and return the top entities. Something like the code below:

``` java
return map.values().stream()
    .sorted(SongEntity.ORDER_BY_SALIENCE_DESCENDING)
    .limit(MAX_ENTITIES)
    .collect(Collectors.toList());
```

src/main/java/com/google/alpollo/model/SongEntity.java
amaurymedeiros: Here you can use `Double.compare()` instead...

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -62,10 +68,25 @@ public boolean equals(Object obj) {
   @Override
   public int hashCode() {
     int hash = 7;
-    hash = 31 * hash + (int) salience;
```
amaurymedeiros, line 65: ... and here you can use `Double.doubleToLongBits()`. That's what `Double.compare()` uses to compare the numbers. Since in both `equals` and `hashCode` you will be comparing the floating-point representation of the numbers, you'll achieve the consistency Jerome mentioned above.

image

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -62,10 +68,25 @@ public boolean equals(Object obj) {
...
+  public Set<String> getType() {
+    return type;
+  }
+
+  public double getSalience() {
+    return salience;
+  }
+
+  public void setSalience(double newSalience) {
+    this.salience = newSalience;
```
amaurymedeiros: nit/optional: when the parameter (i.e. `newSalience`) name is different than the field (i.e. `salience`) you can omit `this`.  So here you could have `salience = newSalience`. If the parameter was also called `salience`, you'd need `this` to differentiate them, i.e. `this.salience = salience;`.

src/main/java/com/google/alpollo/model/SongEntity.java
RazvanBudacaTechquarter: I don't really understand how you want to use `Double.compare()` here.

src/main/java/com/google/alpollo/AnalysisHelper.java
RazvanBudacaTechquarter: Since we merged the 2 methods, this is not needed anymore right?

src/main/java/com/google/alpollo/model/SongEntity.java
amaurymedeiros: `Double.compare(salience, entity.salience)` will return 0 if the numbers are numerically equal.

src/main/java/com/google/alpollo/AnalysisHelper.java
amaurymedeiros: Yep. If you accepted the suggestion below, then this method is irrelevant. Make sure to document the other one if what it does is not obvious from the name.

src/main/java/com/google/alpollo/helpers/AnalysisHelper.java
```diff
@@ -128,4 +127,27 @@ public static String getInterpretation(double score, double magnitude) {
 
     return MIXED;
   }
+
+  public static List<SongEntity> getFilteredTop10(List<SongEntity> list) {
```
amaurymedeiros: nit: maybe change to `getFilteredTopSongs`, in case `MAX_ENTITIES` values change in the future.

src/main/java/com/google/alpollo/model/SongEntity.java
```diff
@@ -17,15 +20,18 @@
...
    * All types can be seen here: 
    * https://cloud.google.com/natural-language/docs/reference/rest/v1/Entity#type
+   * 
+   * Each entity will have multiple, non-duplicate types.
    */ 
-  @Index private String type;
+  private Set<String> type;
   /** Some entities might have a wikipedia link attached to them. */
   @Index private String wikiLink;
+  private static final double TOLERANCE = 0.02;
```
amaurymedeiros: you can probably remove this now

amaurymedeiros (approved):

PR #51, opened 2020-09-08
Save analysis only for top10 songs

sashapff:


jeinygroove (approved):

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -9,13 +9,14 @@
   private static final int TOP_SIZE = 10;
 
   /** Save request song to database and increment the search counter. */
```
amaurymedeiros: nit: add to the doc that we return the `SongCounter` in the end

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -25,7 +26,17 @@ public static void saveSongRequest(Song song) {
 
   /** Save analysis info to database. */
   public static void saveAnalysisInfo(AnalysisInfo info) {
-    OfyService.ofy().save().entity(info).now();
+    // Song that we want to save
+    SongCounter newSongCounter = saveSongRequest(info.getSong());
+
+    List<SongCounter> songCounters = topSongs();
+    // Song that we already have in database
+    SongCounter oldSongCounter = songCounters.get(songCounters.size() - 1);
```
amaurymedeiros: If I understood correctly, this is covering the case where the current song was about to enter the ranking in the last position. Is that right? In this case, I think we can expand the documentation to be more explicit. And maybe `oldSongCounter` could be named `lessSearchedSongFromTop`, or something like that.

jgaillard85 (approved):

PR #52, opened 2020-09-09
Display top song info

jeinygroove:
Now rating list is clickable, it opens song info page for this song, yay!

TODO: if SongInfo with this id doesn't exist, show the error


sashapff (approved):

RazvanBudacaTechquarter (approved):

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -25,7 +25,7 @@ public static SongCounter saveSongRequest(Song song) {
   public static List<SongCounter> topSongs() {
     return OfyService.ofy().load().type(SongCounter.class).order("-searchCounter").limit(TOP_SIZE).list();
   }
-
+
```
jgaillard85: remove whitespaces

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -34,9 +34,10 @@ public static void saveAnalysisInfo(AnalysisInfo info) {
     List<SongCounter> songCounters = topSongs();
     // Song that we already have in database and it is on the last place in the top of songs
     SongCounter lessSearchedSongFromTop = songCounters.get(songCounters.size() - 1);
-
-    if (lessSearchedSongFromTop.getSearchCounter() < newSong.getSearchCounter()) {
+
```
jgaillard85: remove whitespaces

src/main/java/com/google/alpollo/SongDataBase.java
```diff
@@ -34,9 +34,10 @@ public static void saveAnalysisInfo(AnalysisInfo info) {
     List<SongCounter> songCounters = topSongs();
     // Song that we already have in database and it is on the last place in the top of songs
     SongCounter lessSearchedSongFromTop = songCounters.get(songCounters.size() - 1);
-
-    if (lessSearchedSongFromTop.getSearchCounter() < newSong.getSearchCounter()) {
+    
+    if (lessSearchedSongFromTop.getSearchCounter() < newSong.getSearchCounter() || songCounters.size() <= TOP_SIZE) {
       OfyService.ofy().delete().type(AnalysisInfo.class).id(lessSearchedSongFromTop.getSong().id());
+      info.setId();
```
jgaillard85: Why is that necessary? It looks like the id is set at the creation of the AnalysisInfo object

jeinygroove: Because when we send song info object from the frontend it doesn't have id field, that's why it'll be set to null, that's why we should set it manually (make it equal to the id of his SongCounter as well)

jgaillard85: Ok, maybe add a comment to explain the point of this call. Also the name setId is not great because it looks like it is a setter for id, but it takes no argument. Maybe something like initialize, or generateId would be more appropriate

jgaillard85 (approved):

PR #53, opened 2020-09-14
Maven workflow for build check

jeinygroove:
Better now than never :) This workflow will check that pushed changes build successfully. It skips tests, because some of them require to be run on Google Compute Engine

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,24 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


RazvanBudacaTechquarter (approved):

.github/workflows/maven.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Build with Maven
+      run: mvn -B package --file pom.xml -DskipTests
```
amaurymedeiros, line 20: Why are we skipping the tests? Shouldn't we block a PR if it make the tests fail?

jeinygroove: some of the tests work only from Google Compute Engine because of some authorization things, also they require sensitive data and I'm not sure that github can somehow access it without adding this data to repo

amaurymedeiros: Fair enough. For now, let's just agree on running these tests locally before merging the PRs then.

amaurymedeiros (approved):

.github/workflows/maven.yml
jeinygroove: Sure!

Issue #30, opened 2020-08-27
Centralise Material UI components with changed styles.


jeinygroove:
Move `<CircularProgress style={{ color: "black" }} />` and all similar things to the separate file

PR #57, opened 2020-09-15
Text to speech

jeinygroove:
Give users the possibility to listen to the text of lyrics. It can be helpful if they don't want/can't read it or if they want to check the pronunciation of it (for non native speakers).


src/main/java/com/google/alpollo/TextToSpeechServlet.java
```diff
@@ -0,0 +1,35 @@
...
+
+      if (audioContents == null) {
+        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
+            "Couldn't generate audio from the text.");
+        return;
+      }
+
+      byte[] audio = audioContents.toByteArray();
+      response.setContentType("application/json;");
+      response.getWriter().println(new Gson().toJson(audio));
```
RazvanBudacaTechquarter: Why are you making a new Gson object here? You already declared one.

src/main/java/com/google/alpollo/TextToSpeechServlet.java
```diff
@@ -0,0 +1,35 @@
...
+    } catch (Exception e) {
```
RazvanBudacaTechquarter: Is it ok to catch all exceptions instead of specifying each type of exception?

jeinygroove: Yes, since we should not throw any exceptions from the servlet

RazvanBudacaTechquarter: Well not throw them, but specifiy which exceptions you're trying to catch.

RazvanBudacaTechquarter (approved):

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,68 @@
...
+   */
+  public static ByteString convertTextToSpeech(String text, String projectID) throws IOException {
+    createTextToSpeechSettings(projectID);
+
+    try (TextToSpeechClient textToSpeechClient = TextToSpeechClient.create(settings)) {
+      SynthesisInput input = SynthesisInput.newBuilder().setText(text).build();
+      final String languageCode = DetectLanguageService.detectLanguage(text, projectID);
+
+      if (languageCode == null) {
+        throw new IllegalArgumentException("Can't detect language of this text.");
```
amaurymedeiros: We usually throw `IllegalArgumentException` when the arguments passed to the function are invalid. For instance, if you passed a `null` `text` and you didn't support it. Let's throw `IllegalStateException` instead.

src/main/java/com/google/alpollo/DetectLanguageService.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.cloud.translate.v3.DetectLanguageResponse;
+import com.google.cloud.translate.v3.DetectedLanguage;
+import com.google.cloud.translate.v3.LocationName;
+import com.google.cloud.translate.v3.TranslationServiceClient;
+import com.google.cloud.translate.v3.TranslationServiceSettings;
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import java.io.IOException;
+import java.util.List;
+
+public class DetectLanguageService {
```
amaurymedeiros: Let's add a javadoc to the class defining it's main objective.

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * Generates speech audio file from the text.
+   *
+   * @param text the text for convertion
+   * @return ByteString of audio; {@code null} if something went wrong
+   */
+  public static ByteString convertTextToSpeech(String text, String projectID) throws IOException {
+    createTextToSpeechSettings(projectID);
```
amaurymedeiros: It's a little bit odd to have this "set up" method having to being called here. What do you think about transforming `createTextToSpeechSettings` into a `createTextToSpeechClient` method instead? Then you wouldn't need this line and below you'd have `try (TextToSpeechClient textToSpeechClient = createTextToSpeechSettings(projectID))`.

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,68 @@
...
+
+    try (TextToSpeechClient textToSpeechClient = TextToSpeechClient.create(settings)) {
+      SynthesisInput input = SynthesisInput.newBuilder().setText(text).build();
```
amaurymedeiros: `input` is only used way below, so it's better to move it closer to where it's actually needed.

src/main/java/com/google/alpollo/TextToSpeechService.java
amaurymedeiros: also, nit: We know the text is a lyrics, so the exception message could be explicit: "... of this lyrics."

amaurymedeiros: Also, have you considered not throwing an exception at all? If we can't detect the language we could, for instance, fall back to English.

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,68 @@
...
+          AudioConfig.newBuilder()
+              .setAudioEncoding(AudioEncoding.MP3)
+              .build();
+
+      SynthesizeSpeechResponse response =
+          textToSpeechClient.synthesizeSpeech(input, voice, audioConfig);
+
+      return response.getAudioContent();
+    } catch (IOException e) {
+      return null;
```
amaurymedeiros: When we catch an exception and don't log it, we should make it clear in the javadoc, so the callers know this method returns null in certain situations. They need to be aware of when exactly to expect a null return to react accordingly.

src/main/webapp/src/components/lyrics.js
```diff
@@ -1,15 +1,77 @@
...
  * Displays search component for song lyrics.
  */
 class Lyrics extends React.Component {
   constructor(props) {
     super(props);
+
+    this.state = {
+      speechUrl: undefined,
+      isLoading: false,
+      error: null,
```
amaurymedeiros, line 40: Why are we initializing `speechUrl` as `undefined` and `error` as `null`? Shouldn't we use the same value?

src/main/webapp/src/components/lyrics.js
```diff
@@ -20,14 +82,57 @@ class Lyrics extends React.Component {
...
+      <div className={classes.root}>
+        <Typography variant="h4">Lyrics</Typography>
+        <div className={classes.audio}>
+          {AudioPlayerStatus}
+        </div>
+        <p style={{ whiteSpace: "pre-wrap" }}>{lyrics}</p>
+      </div>);};
+
+    if (this.state.error) {
+      console.log(this.state.error.message)
```
amaurymedeiros: Remove logging

src/main/webapp/src/components/lyrics.js
```diff
@@ -1,15 +1,77 @@
...
+});
+
+const StyledLinearProgress = withStyles({
+  root: {
+    backgroundColor: lighten('#000000', 0.5), 
+  },
+  bar: {
+    background: '#000000',
+  },
+})(LinearProgress);
```
amaurymedeiros, line 28: I feel it's a little bit more organized to keep the styles in a CSS file and import them here. So below you'd have something like `<LinearProgress className="StyledLinearProgress" />`.

However, from what I've found online, it's a matter of personal preference, so feel free to keep using inline styles.

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,68 @@
+package com.google.alpollo;
+
+import com.google.cloud.translate.Detection;
+import com.google.api.gax.rpc.FixedHeaderProvider;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.cloud.texttospeech.v1.*;
+import com.google.protobuf.ByteString;
+import java.io.IOException;
+
+public class TextToSpeechService {
```
amaurymedeiros: Add a javadoc

src/main/java/com/google/alpollo/DetectLanguageService.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Generates speech audio file from the text.
+   *
+   * @param text the text for convertion
+   * @return ByteString of audio; {@code null} if something went wrong
+   */
+  public static String detectLanguage(String text, String projectID) throws IOException {
+    createTranslationServiceSettings(projectID);
```
amaurymedeiros: Same comment from the next file. Consider having a `createTranslationServiceClient` method instead and calling it below.

src/main/java/com/google/alpollo/DetectLanguageService.java
```diff
@@ -0,0 +1,65 @@
...
+        DetectLanguageRequest.newBuilder()
+            .setParent(parent.toString())
+            .setMimeType("text/plain")
+            .setContent(text)
+            .build();
+
+      List<DetectedLanguage> response = client.detectLanguage(request).getLanguagesList();
+
+      /* Return null if language wasn't detected or 
+         return the code of the language with the largest confidence. */
```
amaurymedeiros: Multi-line comments should use `//` or have a `*` marking the start of each new line. See the [style guide](https://google.github.io/styleguide/javaguide.html#s4.8.6-comments) for reference.

src/main/webapp/src/components/lyrics.js
jeinygroove: I have an issue #30, so all such things will be moved to the separate files

src/main/webapp/src/components/lyrics.js
jeinygroove: `undefined` means that we just haven't defined/received `speechUrl` yet, but `null` means there's no error

src/main/java/com/google/alpollo/TextToSpeechService.java
jeinygroove: Good idea

src/main/webapp/src/components/lyrics.js
jeinygroove: Why? I have it here so we'll know what exactly went wrong

src/main/webapp/src/components/lyrics.js
amaurymedeiros: Gotcha

src/main/webapp/src/components/lyrics.js
amaurymedeiros: SGTM

amaurymedeiros (commented):
LGTM

src/main/webapp/src/components/lyrics.js
amaurymedeiros: I guess that's OK for debugging, but keep in mind it can cause errors: https://stackoverflow.com/questions/1114187/bad-idea-to-leave-console-log-calls-in-your-production-javascript-code

Also, if it's an error, you can use [console.error](https://developer.mozilla.org/en-US/docs/Web/API/Console/error) instead.

src/main/java/com/google/alpollo/DetectLanguageService.java
```diff
@@ -0,0 +1,67 @@
...
+
+    return TranslationServiceClient.create(settings);
+  }
+
+  /**
+   * Generates speech audio file from the text.
+   *
+   * @param text the text for convertion
+   * @return ByteString of audio; {@code null} if something went wrong
+   */
```
jgaillard85: This documentation does not match the method

src/main/java/com/google/alpollo/DetectLanguageService.java
```diff
@@ -0,0 +1,67 @@
...
+  public static String detectLanguage(String text, String projectID) throws IOException {
+    try (TranslationServiceClient client = createTranslationServiceClient(projectID);) {
```
jgaillard85: unnecessary ';'

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.IOException;
+
+/**
+ * Contains methods for converting text to speech.
+ */
+public class TextToSpeechService {
+  private static TextToSpeechSettings settings;
+  private static final String defaultLanguageCode = "en-US";
+
+  private TextToSpeechService() {};
```
jgaillard85: unnecessary ';'

src/main/java/com/google/alpollo/TextToSpeechService.java
```diff
@@ -0,0 +1,70 @@
...
+import com.google.cloud.texttospeech.v1.*;
+import com.google.protobuf.ByteString;
+import java.io.IOException;
+
+/**
+ * Contains methods for converting text to speech.
+ */
+public class TextToSpeechService {
+  private static TextToSpeechSettings settings;
+  private static final String defaultLanguageCode = "en-US";
```
jgaillard85: Constant fields should have their name fully capitalized

jgaillard85 (approved):

src/main/java/com/google/alpollo/TextToSpeechService.java
jeinygroove: forgot sometimes :(

PR #58, opened 2020-09-16
Autocomplete artist and song name

RazvanBudacaTechquarter:
Created a servlet where the client can send a post request with the current artist name in it's body and it will return a list of the first 5 possible results.
When you search for "Eminem", it's actually the third result. wat.
image


sashapff (approved):

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+      artistName = gson.fromJson(request.getReader(), String.class);
+
+      HttpTransport httpTransport = new NetHttpTransport();
+      HttpRequestFactory requestFactory = httpTransport.createRequestFactory();
+
+      JSONParser parser = new JSONParser();
+
+      GenericUrl url = new GenericUrl("https://kgsearch.googleapis.com/v1/entities:search");
+      url.put("query", artistName);
+      url.put("limit", "5");
```
jeinygroove: Can you move all this magic constants to variables? not the name of parameters, but values: 

```
private static final int LIMIT = 5;
```

and so on

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+      url.put("types", Arrays.asList("MusicGroup", "Person"));
+      url.put("indent", "true");
+      url.put("key", ConfigHelper.getSensitiveData(this.getServletContext(), ConfigHelper.SENSITIVE_DATA.API_KEY));
+
+      HttpRequest autocompleteRequest = requestFactory.buildGetRequest(url);
+      HttpResponse autocompleteResponse = autocompleteRequest.execute();
+      JSONObject responseObject = (JSONObject) parser.parse(autocompleteResponse.parseAsString());
```
jeinygroove: why you use JSONParser here, can you use gson instead?

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+      JSONArray elements = (JSONArray) responseObject.get("itemListElement");
+
+      List<String> artists = new ArrayList<>();
+      for (Object element : elements) {
+        artists.add(JsonPath.read(element, "$.result.name").toString());
```
jeinygroove: can you, please, explain what's happening here?

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -206,4 +220,60 @@ public void sentimentIsMixed() {
...
+  @Test
+  public void foundArtistWithWrongName() throws IOException {
+    when(request.getReader()).thenReturn(
+        new BufferedReader(new StringReader(gson.toJson(WRONG_ARTIST_NAME))));
+    artistServletUnderTest.doPost(request, response);
+    String responseString = responseWriter.toString();
+
+    List<String> actual = gson.fromJson(responseString, new TypeToken<List<String>>(){}.getType());
+    List<String> expected = Arrays.asList();
+
```
jeinygroove, line 272: nit: empty line

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -206,4 +220,60 @@ public void sentimentIsMixed() {
...
+  @Test
+  public void foundArtistWithIncompleteName() throws IOException {
+    when(request.getReader()).thenReturn(
+        new BufferedReader(new StringReader(gson.toJson(INCOMPLETE_ARTIST_NAME))));
+    artistServletUnderTest.doPost(request, response);
+    String responseString = responseWriter.toString();
+
+    List<String> actual = gson.fromJson(responseString, new TypeToken<List<String>>(){}.getType());
+    List<String> expected = Arrays.asList("Dr. Dre", "Michael Clarke", "Eminem", "Emine Erdoğan", "Emine Gülbahar Hatun");
+
```
jeinygroove: nit: empty line

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -206,4 +220,60 @@ public void sentimentIsMixed() {
...
+  @Test
+  public void artistIsABand() throws IOException {
+    when(request.getReader()).thenReturn(
+        new BufferedReader(new StringReader(gson.toJson(INCOMPLETE_BAND_NAME))));
+    artistServletUnderTest.doPost(request, response);
+    String responseString = responseWriter.toString();
+
+    List<String> actual = gson.fromJson(responseString, new TypeToken<List<String>>(){}.getType());
+    List<String> expected = Arrays.asList("Rammstein", "Otthein Rammstedt", "Rammstedt");
+
```
jeinygroove, line 285: nit: empty line

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+
+      HttpTransport httpTransport = new NetHttpTransport();
+      HttpRequestFactory requestFactory = httpTransport.createRequestFactory();
+
+      JSONParser parser = new JSONParser();
+
+      GenericUrl url = new GenericUrl("https://kgsearch.googleapis.com/v1/entities:search");
+      url.put("query", artistName);
+      url.put("limit", "5");
+      url.put("types", Arrays.asList("MusicGroup", "Person"));
```
jeinygroove: why do we need "Person" here? does there exist something like "Musician" or smth like that for musicians like `David Bowie`, `Elvis Presley`, etc.?

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+      url.put("indent", "true");
```
jeinygroove: what this parameter means?

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
```diff
@@ -0,0 +1,73 @@
...
+
+      List<String> artists = new ArrayList<>();
+      for (Object element : elements) {
+        artists.add(JsonPath.read(element, "$.result.name").toString());
+      }
+
+      String json = gson.toJson(artists);
+      response.setContentType("application/json");
+      response.getWriter().println(json);
+    } catch (JsonSyntaxException | JsonIOException | IOException | ParseException autoException) {
```
jeinygroove: I think you can replace it with `Exception`

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -206,4 +220,60 @@ public void sentimentIsMixed() {
...
+  @Test
+  public void foundArtistWithFullName() throws IOException {
+    when(request.getReader()).thenReturn(
+        new BufferedReader(new StringReader(gson.toJson(FULL_ARTIST_NAME))));
+    artistServletUnderTest.doPost(request, response);
+    String responseString = responseWriter.toString();
+
+    List<String> actual = gson.fromJson(responseString, new TypeToken<List<String>>(){}.getType());
+    List<String> expected = Arrays.asList("Dr. Dre", "Michael Clarke", "Eminem", "Bad Meets Evil", "Vincent Vinel");
+
```
jeinygroove: nit: empty line

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
RazvanBudacaTechquarter: I tried to use gson and it doesn't work, so I used the same library the google documentation uses

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: But I like those empty lines, they separate the initializations from the actual execution and then the execution from the assertion

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,77 @@
...
+   * Making a POST request to this servlet with a search string and the types searched for
+   * as parameters will make a call to the Knowledge Graph Search API and it will return a list of
+   * possible search results.
+   * 
+   * The artist's name doesn't necessarily have to be complete
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String searchString;
+    String types;
```
jeinygroove: I think you didn't understand what I meant, client won't send to you types of entities, because it's not familiar with internal api and it shouldn't. I meant that type can be "artist" or "song"

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
RazvanBudacaTechquarter: So each json element has a path to `result.name` where you can find the name of the result. I go through each element and store the name of each result.
image

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
RazvanBudacaTechquarter: Nope, just Person. You can see all types of entities [here](https://developers.google.com/knowledge-graph#knowledge_graph_entities)

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
RazvanBudacaTechquarter: Removed it, apparently it's useless for us.

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
RazvanBudacaTechquarter: I thought it's better to mention which exceptions we're trying to catch.

src/test/java/com/google/alpollo/AnalysisTest.java
jeinygroove: you can have one empty line, of course, but why two :D

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
jeinygroove: gson has JsonParser as well (it was deprecated, but have a look at the suggested alternatives [here](https://javadoc.io/doc/com.google.code.gson/gson/latest/com.google.gson/com/google/gson/JsonParser.html)). Have you tried this?

RazvanBudacaTechquarter: JsonPath from jayway apparently doesn't like with the jsons created by gson. This is what I'm getting if I use gson:
`Expected to find an object with property ['result'] in path $ but found 'com.google.gson.JsonObject'. This is not a json object according to the JsonProvider: 'com.jayway.jsonpath.spi.json.JsonSmartJsonProvider'.`

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: I see only one empty line between each of the three code blocks.

src/main/java/com/google/alpollo/servlets/AutocompleteArtistServlet.java
jeinygroove: ok, let's skip it for now then

jeinygroove (approved):

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,90 @@
...
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+@WebServlet("/autocomplete")
+public class AutocompleteServlet extends HttpServlet {
+  private final Gson gson = new Gson();
+  private final static String SEARCH_STRING = "searchString";
+  private final static String TYPE = "type";
+  private final static String LIMIT = "10";
+  private List<String> typeList;
```
jgaillard85: It looks like this could be a local variable

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,90 @@
...
+
+      HttpRequest autocompleteRequest = requestFactory.buildGetRequest(url);
+      HttpResponse autocompleteResponse = autocompleteRequest.execute();
+      JSONObject responseObject = (JSONObject) parser.parse(autocompleteResponse.parseAsString());
+      JSONArray elements = (JSONArray) responseObject.get("itemListElement");
+
+      List<String> results = new ArrayList<>();
+      for (Object element : elements) {
+        results.add(JsonPath.read(element, "$.result.name").toString());
+      }
```
jgaillard85: Can this not done with Gson, instead of adding new dependencies?

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * Making a POST request to this servlet with a search string and the type searched for
+   * as parameters will make a call to the Knowledge Graph Search API and it will return a list of
+   * possible search results.
+   */
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String searchString;
+    SearchType type;
```
jgaillard85: Those two variables don't seem to need to be defined outside of the try

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -214,6 +234,85 @@ public void sentimentIsMixed() {
...
+  @Test
+  public void autocompleteBadTypeRequest() throws IOException {
+    when(request.getParameter(SEARCH_STRING)).thenReturn(INCOMPLETE_SONG_NAME);
+    when(request.getParameter(TYPE)).thenReturn(PAINTING);
+    autocompleteServletUnderTest.doPost(request, response);
+    
+    verify(response).sendError(
+        HttpServletResponse.SC_BAD_REQUEST,
+        "Type not supported.");
+  }
```
jgaillard85: This test is not very useful as you are the one creating the request, so that should never happen

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
RazvanBudacaTechquarter: JsonPath from jayway apparently doesn't like to work with the jsons created by gson. This is what I'm getting if I use gson:
`Expected to find an object with property ['result'] in path $ but found 'com.google.gson.JsonObject'. This is not a json object according to the JsonProvider: 'com.jayway.jsonpath.spi.json.JsonSmartJsonProvider'.`

src/test/java/com/google/alpollo/AnalysisTest.java
RazvanBudacaTechquarter: Should I just delete it?

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
jgaillard85: It looks like you might be able to have JsonPath work with Gson by using a custom configuration: https://github.com/json-path/JsonPath#tweaking-configuration

src/test/java/com/google/alpollo/AnalysisTest.java
jgaillard85: Yes, I don't think it brings any value

src/test/java/com/google/alpollo/AnalysisTest.java
```diff
@@ -68,6 +70,17 @@
...
+  private static final String FULL_ARTIST_NAME = "Eminem";
+  private static final String INCOMPLETE_ARTIST_NAME = "Emine";
+  private static final String WRONG_ARTIST_NAME = "Emnem";
+  private static final String INCOMPLETE_BAND_NAME = "Rammste";
+  private static final String INCOMPLETE_SONG_NAME = "master o";
+  private static final String SEARCH_STRING = "searchString";
+  private static final String TYPE = "type";
+  private static final String ARTIST = "ARTIST";
+  private static final String SONG = "SONG";
+  private static final String PAINTING = "PAINTING";
```
jgaillard85: That's probably not needed any more

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+
+      GenericUrl url = new GenericUrl("https://kgsearch.googleapis.com/v1/entities:search");
+      url.put("query", searchString);
+      url.put("limit", LIMIT);
+      url.put("types", typeList);
+      url.put("key", ConfigHelper.getSensitiveData(this.getServletContext(), ConfigHelper.SENSITIVE_DATA.API_KEY));
+
+      HttpRequest autocompleteRequest = requestFactory.buildGetRequest(url);
+      HttpResponse autocompleteResponse = autocompleteRequest.execute();
+      configJsonPathWithGson();
```
jgaillard85: Do you need to do that for every request?

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+      JsonObject responseObject = (JsonObject) JsonParser.parseString(autocompleteResponse.parseAsString());
+      JsonArray elements = (JsonArray) responseObject.get("itemListElement");
```
jgaillard85, line 83: The idea of using JsonPath with Gson was to be able to use Gson here and remove the need for the json-simple dependency

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
RazvanBudacaTechquarter: I don't see any other place I could do it. Do you have any ideas?

jgaillard85: Maybe this can be done in the constructor of this class, or during static initialization of the class

src/main/java/com/google/alpollo/servlets/AutocompleteServlet.java
```diff
@@ -0,0 +1,118 @@
...
+      response.setContentType("application/json");
+      response.getWriter().println(json);
+    } catch (JsonSyntaxException | JsonIOException | IOException autoException) {
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, autoException.getMessage());
+    } catch (IllegalArgumentException enumException) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Type not supported.");
+    }
+  }
+
+  private void configJsonPathWithGson() {
```
jgaillard85: Make this method static

jgaillard85 (approved):

PR #59, opened 2020-09-17
Add autocomplete field for artist name

jeinygroove:


sashapff (approved):

src/main/webapp/src/components/autocompleteTextField.js
```diff
@@ -0,0 +1,96 @@
...
+    "& .MuiInput-underline:after": {
+      borderBottomColor: "#733F94",
+    },
+    "& .MuiOutlinedInput-root": {
+      "&.Mui-focused fieldset": {
+        borderColor: "#733F94",
+      },
+    },
+  },
+})(TextField);
```
amaurymedeiros, line 26: Up to you, but I think it'd be more organized to keep the styles in a css file

src/main/webapp/src/components/autocompleteTextField.js
```diff
@@ -0,0 +1,96 @@
...
+
+export default function AutocompleteTextField(props) {
+  const [value, setValue] = React.useState(props.value);
+  const [options, setOptions] = React.useState([]);
+
+  const fetch = React.useMemo(
+    () =>
+      throttle((request, callback) => {
+        axios
+          .post(`/autocomplete?type=${props.type}&searchString=${request}`)
```
amaurymedeiros: In POST requests, we don't pass the parameters in the URL. For `axios`, I'm pretty sure the syntax would be `post('/autocomplete', { type: props.type, searchString: request })`, but maybe you need to adjust it slightly.

src/main/webapp/src/components/autocompleteTextField.js
```diff
@@ -0,0 +1,96 @@
...
+      }, 200),
+    [],
+  );
+
+  React.useEffect(() => {
+    let active = true;
+
+    if (value === '') {
+      return undefined;
+    }
```
amaurymedeiros: You can move this before `let active = true;`. Also, probably `return;` would be enough. Please add a comment explaining why we're returning early when value is empty.

src/main/webapp/src/components/autocompleteTextField.js
```diff
@@ -0,0 +1,96 @@
...
+
+    fetch(value, (results) => {
+      if (active) {
```
amaurymedeiros, line 52: Why do you need the `active` flag at all?

src/main/webapp/src/components/autocompleteTextField.js
jeinygroove: Will be done in #30

src/main/webapp/src/components/autocompleteTextField.js
jeinygroove: Because we have async call in it and we need to control that we change state of smth that still exists. Longer explanation can be found [here](https://stackoverflow.com/questions/62453435/why-do-we-need-to-return-a-function-in-the-react-hook).

jgaillard85 (approved):

